PR #25, opened 2017-06-29
New ant_metrics module

jsdillon:


jsdillon:
Next up before this is ready are unit tests, though I'd like feedback on functionality and interchange formats too. Let's discuss in tomorrow's commissioning meeting.

jsdillon:
@adampbeardsley I think this is ready to be reviewed and hopefully merged.

hera_qm/ant_metrics.py
```diff
@@ -0,0 +1,394 @@
...
+            worstDeadCutRatio = np.abs(deadMetrics[worstDeadAnt])/deadCut
+
+            # Most likely cross-polarized antenna 
+            crossMetrics = average_abs_metrics(self.allModzScores[-1]['meanVijXPol'],
+                                               self.allModzScores[-1]['redCorrXPol'])
+            worstCrossAnt = max(crossMetrics, key=crossMetrics.get)
+            worstCrossCutRatio = np.abs(crossMetrics[worstCrossAnt])/crossCut
+
+            # Find the single worst antenna, remove it, log it, and run again
+            if worstCrossCutRatio >= worstDeadCutRatio and worstCrossCutRatio >= 1.0:
```
adampbeardsley: This doesn't evaluate to true for any of the tests. We should make sure it gets covered, considering it's a significant piece of code.

hera_qm/tests/test_ant_metrics.py
```diff
@@ -0,0 +1,148 @@
...
+                 (72, 105), (22, 9), (81, 20), (112, 22), (89, 31), (97, 81), (105, 88)], [(9, 112), 
+                 (20, 97), (53, 81), (31, 10), (80, 20), (64, 22), (96, 43), (88, 72), (104, 89)], 
+                 [(9, 81), (22, 97), (31, 43), (53, 89), (105, 112), (20, 10), (64, 20), (88, 22), 
+                 (80, 31), (104, 65)], [(43, 72), (65, 105), (96, 88)], [(31, 97), (53, 112), 
+                 (64, 72), (96, 10), (80, 22), (104, 81)], [(10, 88), (43, 64)], [(9, 97), (64, 81), 
+                 (80, 89), (88, 112), (53, 10), (104, 43)]]
+        self.am = ant_metrics.Antenna_Metrics(dataFileList, reds, fileformat='miriad')
+        with self.assertRaises(KeyError):
+            self.am.save_antenna_metrics(DATA_PATH + '/test_output/ant_metrics_output.json')
+        self.am.iterative_antenna_metrics_and_flagging()
```
adampbeardsley: This setup is organized kind of funny. There shouldn't be an assert inside the setup, and I don't think the iterative function needs to be run (e.g. test_nPos doesn't need it.) I think it would make more sense to define `self.reds` and instantiate `self.am` in the tests when needed. Alternatively, is it possible to define an __init__ for this function that would run the tests just once? the setup gets run for each test.

hera_qm/ant_metrics.py
```diff
@@ -0,0 +1,394 @@
...
+    for each antenna, which is the metrics, minus the median, divided by the median absolute deviation.'''
+    zscores = {}    
+    antpols = set([key[1] for key in metric.keys()])
+    for antpol in antpols:            
+        values = np.array([val for key,val in metric.items() if key[1]==antpol])
+        median = np.nanmedian(values)
+        medAbsDev = np.nanmedian(np.abs(values - median))
+        for key,val in metric.items(): 
+            if key[1]==antpol:
+                zscores[key] = 0.6745*(val - median) / medAbsDev
```
jsdillon: I've assigned myself an issue: #27 

This should wait until we're comfortable with the operation and effectiveness of the metrics.

hera_qm/ant_metrics.py
```diff
@@ -0,0 +1,394 @@
...
+import json
+
+
+#######################################################################
+#Low level functionality that is potentially reusable
+#######################################################################
+
+
+def per_antenna_modified_z_scores(metric):
+    '''For a given metric, stored as a (ant,antpol) dictonary, computes the per-pol modified z-score
```
jsdillon: I use this format for "simple," self-explanatory functions and the Arguments/Returns format for more complicated ones. Let me know if you think we definitely need to change it. 

I personally find that huge comment blocks makes the code less readable because it separates the function definition from the rest of the function.

hera_qm/ant_metrics.py
```diff
@@ -0,0 +1,394 @@
...
+    '''
+    
+    absVijMean = {(ant,antpol):0.0 for ant in ants for antpol in antpols if (ant,antpol) not in xants}
+    visCounts = deepcopy(absVijMean)
+    for (i,j) in bls:
+        if i != j:
+            for pol in pols:
+                for ant, antpol in zip((i,j), pol):
+                    if (ant,antpol) not in xants:
+                        absVijMean[(ant,antpol)] += np.abs(data.get_data(i,j,pol))
```
jsdillon: Good idea. I'll do that.

hera_qm/ant_metrics.py
```diff
@@ -0,0 +1,394 @@
...
+                        visCounts[(ant,antpol)] += 1
+    timeFreqMeans = {key: np.nanmean(absVijMean[key] / visCounts[key]) for key in absVijMean.keys()}
+
+    if rawMetric: 
+        return timeFreqMeans
+    else: 
+        return per_antenna_modified_z_scores(timeFreqMeans)
+
+def red_corr_metrics(data, pols, antpols, ants, reds, xants=[], rawMetric=False, crossPol=False):
+    '''Calculates the extent to which baselines involving an antenna don't correlated
```
jsdillon: Got it.

hera_qm/ant_metrics.py
```diff
@@ -0,0 +1,394 @@
...
+        allMetricsData = {'xants': str(self.xants)}
+        allMetricsData['ants_removed_as_crossed'] = str(self.crossedAntsRemoved)
+        allMetricsData['ants_removed_as_dead'] = str(self.deadAntsRemoved)
+        allMetricsData['final_metrics'] = str(self.finalMetrics)
+        allMetricsData['all_metrics'] = str(self.allMetrics)
+        allMetricsData['final_mod_z_scores'] = str(self.finalModzScores)
+        allMetricsData['all_mod_z_scores'] = str(self.allModzScores)
+        allMetricsData['removal_iteration'] = str(self.removalIter)
+        allMetricsData['cross_pol_z_cut'] = str(self.crossCut)
+        allMetricsData['dead_ant_z_cut'] = str(self.deadCut)
```
jsdillon: Done.

hera_qm/ant_metrics.py
```diff
@@ -0,0 +1,394 @@
...
+        self.antpols = [antpol.lower() for antpol in self.data.get_feedpols()]
+        self.bls = list(set([(i,j) for (i,j,pol) in self.data.get_antpairpols()]))
+        self.reds = reds
+
+        # For using data containers until pyuvdata gets faster
+        # from hera_cal import firstcal
+        # from hera_qm.datacontainer import DataContainer
+        # self.uvdata = self.data
+        # datapack, flagspack = firstcal.UVData_to_dict([self.uvdata])
+        # self.data = DataContainer(datapack)
```
jsdillon: We can when pyuvdata is as fast as DataContainer or when we delete DataContainer from hera_qm.

hera_qm/ant_metrics.py
```diff
@@ -0,0 +1,394 @@
...
+        if fileformat is 'miriad':
+            self.data.read_miriad(dataFileList)
+        elif fileformat is 'uvfits':
+            self.data.read_uvfits(dataFileList)
+        else:
+            raise ValueError('Unrecognized file format' + str(fileformat))
+        self.ants = self.data.get_ants()
+        self.pols = [pol.lower() for pol in self.data.get_pols()]
+        self.antpols = [antpol.lower() for antpol in self.data.get_feedpols()]
+        self.bls = list(set([(i,j) for (i,j,pol) in self.data.get_antpairpols()]))
```
jsdillon: Ah, I missed that. Ok...I'll use that instead.

hera_qm/ant_metrics.py
```diff
@@ -0,0 +1,394 @@
...
+    return list(xantSet)
+
+
+def antpol_metric_sum_ratio(ants, antpols, crossMetrics, sameMetrics, xants=[]):
+    '''Takes the ratio of two antenna metrics, summed over both polarizations, and creates a new
+    antenna metric with the same value in both polarizations for each antenna.'''
+    crossPolRatio = {}
+    for ant in ants: 
+        if (ant,antpols[0]) not in xants:
+            crossSum = np.sum([crossMetrics[(ant,antpol)] for antpol in antpols])
```
jsdillon: I'll fix it and add a unit test to cover it.

hera_qm/ant_metrics.py
```diff
@@ -0,0 +1,394 @@
...
+    antpols -- List of antenna polarizations (e.g. ['x', 'y'])
+    ants -- List of all antenna indices.
+    bls -- List of tuples of antenna pairs.
+    xants -- list of antennas in the (ant,antpol) format that should be ignored. If, e.g., (81,'y')
+            is excluded, (81,'x') cannot be identified as cross-polarized and will be excluded.
+    rawMetric -- return the raw power ratio instead of the modified z-score
+
+    Returns:
+    mean_Vij_cross_pol_metrics -- a dictionary indexed by (ant,antpol) of the modified z-scores of the  
+            ratio of mean visibilities, (Vxy+Vyx)/(Vxx+Vyy). Results duplicated in both antpols.
```
jsdillon: No idea what happened there...

hera_qm/ant_metrics.py
```diff
@@ -0,0 +1,394 @@
...
+
+class Antenna_Metrics():
+    '''Object for holding relevant visibility data and metadata with interfaces to four 
+    antenna metrics (two for identifying dead antennas, two for identifying cross-polarized ones), 
+    an iterative method for identifying one bad antenna at a time while keeping track of all 
+    metrics, and for writing metrics to a JSON.'''
+    
+    def __init__(self, dataFileList, reds, fileformat='miriad'):
+        '''Arguments:
+        dataFileList -- List of data filenames for the different polarizations
```
jsdillon: Added to the documentation.

hera_qm/ant_metrics.py
adampbeardsley: Ok, I agree simple ones are nice. I think this one's right on the cusp because it's where you define the metric dictionary. I'll leave it up to you.

hera_qm/ant_metrics.py
adampbeardsley: Agreed.

hera_qm/ant_metrics.py
```diff
@@ -0,0 +1,394 @@
...
+        reds -- List of lists of tuples of antenna numbers that make up redundant baseline groups
+        format -- default miriad. Otheqr options: uvfits
```
jsdillon: Done.

hera_qm/ant_metrics.py
```diff
@@ -0,0 +1,394 @@
...
+        # self.data = DataContainer(datapack)
+
+        if len(self.antpols) is not 2 or len(self.pols) is not 4:
+            raise ValueError('Missing polarization information. pols =' + str(self.pols) + ' and antpols = ' + str(self.antpols))
+
+    def mean_Vij_metrics(self, pols=None, xants=[], rawMetric=False):
+        '''Local wrapper for mean_Vij_metrics in hera_qm.ant_metrics module.'''
+
+        if pols is None:
+            pols = self.pols
```
jsdillon: I could go either way on this. `xants` is something that changes during the lifetime of an `Antenna_Metrics` object while pols generally doesn't. I think I'll leave it as is.

hera_qm/ant_metrics.py
jsdillon: Since you wrote that, I added another test that covers this part.

hera_qm/ant_metrics.py
```diff
@@ -0,0 +1,394 @@
...
+    def iterative_antenna_metrics_and_flagging(self, crossCut=5, deadCut=5, verbose=False):
+        '''Runs all four metrics (two for dead antennas two for cross-polarized antennas) and saves
+        the results internally to this this antenna metrics object. 
+        
+        Arguments:
+        crossCut -- Modified z-score cut for most cross-polarized antenna. Default 5 "sigmas".
+        deadCut -- Modified z-score cut for most likely dead antenna. Default 5 "sigmas".
+        '''
+        
+        #Summary statistics
```
jsdillon: Since these summary statistics really only make sense in the context of running all four metrics iteratively, I'm OK with them only getting instantiated here. However, I do see some rationale in letting the user reset them manually, so I'll break that out.

hera_qm/tests/test_ant_metrics.py
jsdillon: I was confused about how setUp works. 

I've reorganized the tests.

jsdillon:
Thanks for your help @adampbeardsley

PR #32, opened 2017-07-22
Ant metrics wrapper

plaplant:
This PR adds a wrapper function for running several ant metrics tests on files passed in. The function is suitable for calling from a script to be part of RTP. Several helper functions were also added to support the wrapper (mostly file system interaction things).

### DIFF
--- a/hera_qm/ant_metrics.py
+++ b/hera_qm/ant_metrics.py
@@ -1,9 +1,13 @@
+from __future__ import print_function, division, absolute_import
 import numpy as np
 import matplotlib.pyplot as plt
-import aipy as a
+import aipy
 from copy import deepcopy
 from pyuvdata import UVData
 import json
+import optparse
+import os
+import re
 from hera_qm.version import hera_qm_version_str
 
 
@@ -398,13 +402,13 @@ class Antenna_Metrics():
                     self.crossedAntsRemoved.append((worstCrossAnt[0], antpol))
                     self.removalIter[(worstCrossAnt[0], antpol)] = n
                     if verbose:
-                        print 'On iteration', n, 'we flag', (worstCrossAnt[0], antpol)
+                        print('On iteration', n, 'we flag', (worstCrossAnt[0], antpol))
             elif worstDeadCutRatio > worstCrossCutRatio and worstDeadCutRatio > 1.0:
                 self.xants.append(worstDeadAnt)
                 self.deadAntsRemoved.append(worstDeadAnt)
                 self.removalIter[worstDeadAnt] = n
                 if verbose:
-                    print 'On iteration', n, 'we flag', worstDeadAnt
+                    print('On iteration', n, 'we flag', worstDeadAnt)
             else:
                 break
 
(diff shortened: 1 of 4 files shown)


plaplant:
@adampbeardsley to build the redundant array info, I have a statement inside the wrapper function `from hera_cal import omni`. Does this create a circular dependency?

hera_qm/ant_metrics.py
```diff
@@ -431,3 +436,155 @@ def save_antenna_metrics(self, metricsJSONFilename):
 
         with open(metricsJSONFilename, 'w') as outfile:
             json.dump(allMetricsData, outfile, indent=4)
+
+# code for running ant_metrics on a file
+def get_metrics_OptionParser(method_name):
```
adampbeardsley: My guess is some of these helper functions (get_metrics_OptionParser, get_pol, generate_fullpol_file_list) will be generally useful, not just for ant_metrics. Would they make sense in a hera_qm_utils module or in the __init__.py file?

hera_qm/ant_metrics.py
```diff
@@ -431,3 +436,155 @@ def save_antenna_metrics(self, metricsJSONFilename):
...
+       None
+
+    The funciton will take in a list of files and options. It will run the
+    sereis of ant metrics tests, and produce a JSON file containing the relevant
+    information. The file list passed in need only contain one of the polarization
+    files for a given JD, and the function will look for the other polarizations
+    in the same folder. If not all four polarizations are found, a warning is
+    generated, since the code assumes all four polarizations are present.
+    """
+    # XXX: does this create a circular dependency if we only do this inside the function?
```
adampbeardsley: Technically I think it does. We could sidestep it by importing omni.aa_to_info directly, which doesn't use hera_qm. Also, Danny is working on an alternative function for getting info. Once that's integrated into hera_cal, we could import that instead.

hera_qm/ant_metrics.py
```diff
@@ -431,3 +436,155 @@ def save_antenna_metrics(self, metricsJSONFilename):
...
+    try:
+        from hera_cal import omni
+    except(ImportError):
+        from nose.plugins.skip import SkipTest
+        raise SkipTest('hera_cal.omni not detected. It must be installed to calculate array info')
+
+    # define polarizations to look for
+    # XXX: This will have to change if file naming convention changes from 'xx', 'xy', etc.
+    pol_list = ['xx', 'xy', 'yx', 'yy']
```
jsdillon: Can't this be generalized using `get_pol`?

hera_qm/ant_metrics.py
```diff
@@ -431,3 +436,155 @@ def save_antenna_metrics(self, metricsJSONFilename):
...
+    if len(files) == 0:
+        raise AssertionError('Please provide a list of visibility files')
+
+    # generate a list of all files to be read in
+    fullpol_file_list = generate_fullpol_file_list(files, pol_list)
+    if len(fullpol_file_list) == 0:
+        raise AssertionError('Could not find all 4 polarizations for any files provided')
+
+    # define freqs
+    # XXX: might want to make this flexible in the future
```
jsdillon: Fair, though all we really need here are the reds.

hera_qm/ant_metrics.py
```diff
@@ -431,3 +436,155 @@ def save_antenna_metrics(self, metricsJSONFilename):
...
+        fname -- name of file (string)
+    Returns:
+        polarization -- polarization label e.g. "xx" (string)
+    """
+    # XXX assumes file naming format
+    # extract just the filename if we're passed a path with periods in it
+    fn = re.findall('zen\.\d{7}\.\d{5}\..*', fname)[0]
+    return fn.split('.')[3]
+
+def generate_fullpol_file_list(files, pol_list):
```
jsdillon: This seems like an ugly way to store the files. To my mind, they should be a list of lists...the outer list loops over JD and the inner list loops over polarization. Then you can just iterate over the outer list and get the list you need as an input to `Antenna_Metrics`.

hera_qm/ant_metrics.py
```diff
@@ -431,3 +436,155 @@ def save_antenna_metrics(self, metricsJSONFilename):
...
+    # process calfile
+    aa = aipy.cal.get_aa(opts.cal, freqs)
+    info = omni.aa_to_info(aa, pols=[pol_list[-1][0]], crosspols=[pol_list[-1]])
+    reds = info.get_reds()
+
+    # do the work
+    num_files = len(fullpol_file_list) // 4
+    for i in range(num_files):
+        file_list = fullpol_file_list[4*i:4*(i+1)]
+        am = Antenna_Metrics(file_list, reds, fileformat='miriad')
```
jsdillon: Might want to make miriad the default of a command line option.

hera_qm/ant_metrics.py
plaplant: I agree, they do seem generally useful. I'm personally not a huge fan of hiding large chunks of code in `__init__.py`, so I'll make a new utils module.

hera_qm/ant_metrics.py
plaplant: @adampbeardsley Okay, I'll import just that function. When Danny's is ready, we can use that one instead.

hera_qm/ant_metrics.py
```diff
@@ -431,3 +436,155 @@ def save_antenna_metrics(self, metricsJSONFilename):
...
+
+# helper functions
+def get_pol(fname):
+    """Strips the filename of a HERA visibility to its polarization
+    Args:
+        fname -- name of file (string)
+    Returns:
+        polarization -- polarization label e.g. "xx" (string)
+    """
+    # XXX assumes file naming format
```
plaplant: Okay, I'll add that

hera_qm/ant_metrics.py
plaplant: Good call, I'll change it to a list of lists.

hera_qm/ant_metrics.py
plaplant: My thinking behind this is that the helper function `generate_fullpol_file_list` needs to know what a "complete set" of polarized visibilities is. I suppose we could have the options include a `-p` option like `omni_run`, and if it's not specified, we default to this pol_list.

hera_qm/ant_metrics.py
plaplant: Okay, if the frequency range doesn't affect the redundancy calculation, then I'll change the comment to something denoting it's just a placeholder frequency range.

hera_qm/ant_metrics.py
plaplant: Will do.

hera_qm/ant_metrics.py
```diff
@@ -431,3 +436,155 @@ def save_antenna_metrics(self, metricsJSONFilename):
...
+    aa = aipy.cal.get_aa(opts.cal, freqs)
+    info = omni.aa_to_info(aa, pols=[pol_list[-1][0]], crosspols=[pol_list[-1]])
+    reds = info.get_reds()
+
+    # do the work
+    num_files = len(fullpol_file_list) // 4
+    for i in range(num_files):
+        file_list = fullpol_file_list[4*i:4*(i+1)]
+        am = Antenna_Metrics(file_list, reds, fileformat='miriad')
+        am.iterative_antenna_metrics_and_flagging(crossCut=opts.crossCut, deadCut=opts.deadCut, verbose=True)
```
plaplant: Fair enough. I'll make it a `-v` flag

hera_qm/ant_metrics.py
jsdillon: Actually, that's a good idea. I like the idea that polarizations are passed (and that you must pass exactly 4 pols made from two antpols) but that this set is the default. Maybe some syntax like `-p xx,xy,yx,yy`.

plaplant: Do the ant_metrics expect that it's 4 polarizations, with 2 auto-pols and 2 cross-pols? If so, I can add a function that checks that's enforced. Yes, that's the syntax for `omni_run`, so I was going to use something similar here

plaplant:
@jsdillon I think this is ready for another look. The only functionality I haven't added yet is asserting that the polarizations specified as a command line option be exactly 4 polarizations long, and have 2 auto-pols and 2 cross-pols. Let me know if you want that check added.

hera_qm/ant_metrics.py
plaplant: I actually made a `-q` flag for quiet operation, i.e., the default is to be verbose.

PR #33, opened 2017-07-25
xrfi clean

jkerrigan:
Cleaned up xrfi/data processing algorithms. Made functions more general by allowing for varying kernel sizes in time/frequency. Expanded coverage tests to account for cases where certain algorithms were known to fail.


hera_qm/tests/test_xrfi.py
```diff
@@ -199,11 +262,13 @@ def rfi_gen():
...
-        self.ans['xrfi_simple'] = (.90, .1)
-
+        self.ans['detrend_deriv'] = [(.83, .83, .83),(.1,.1,.1)]
+        self.ans['detrend_medminfilt'] = [(.2, .3, .3),(.1,.1,.1)]
+        self.ans['xrfi'] = (.7, .1)
+        self.ans['xrfi_simple'] = [(.90, .90),(.1, .1)]
+        self.mode['detrend_deriv'] = [False,False,True]
+        self.mode['detrend_medminfilt'] = [False,False,False]
+        self.mode['watershed'] = [True,True]
 # class TestHERA(Template, unittest.TestCase):
```
adampbeardsley: I know you didn't write these commented tests, but could you look them over and determine if we really want to keep them? If we want them, they should be run. If we don't want them, let's delete.
This is also the only place where `FILES` is used (see above)

hera_qm/tests/test_xrfi.py
```diff
@@ -4,12 +4,14 @@
...
+
 
 np.random.seed(0)
 
 SIZE = 100
 VERBOSE = False
-PLOT = False
+LOT = False
 
 FILES = {
```
adampbeardsley: If possible, please delete this dict. It's only used in commented-out tests (see below). As it stands, these file locations don't even exist.

hera_qm/tests/test_xrfi.py
```diff
@@ -26,22 +28,30 @@ def get_accuracy(f, rfi, verbose=VERBOSE):
...
+    #capo.plot.waterfall(f, mode='lin', mx=10, drng=10)
+    plt.imshow(f,aspect='auto',cmap='jet')
     plt.colorbar()
     plt.show()
 
 
 def plot_result(f, rfi):
-    if not PLOT:
-        return
+    #if not PLOT:
```
adampbeardsley: Why are these lines commented out, along with the calls to plotting in tests? This should all be handled by the PLOT global variable.

hera_qm/tests/test_xrfi.py
```diff
@@ -26,22 +28,30 @@ def get_accuracy(f, rfi, verbose=VERBOSE):
         print '\t Found RFI: %1.3f\n\t False Positive: %1.3f' % (correctly_flagged, false_positive)
     return correctly_flagged, false_positive
 
+def fake_flags():
+    RFI = 10
```
adampbeardsley: This isn't used.

hera_qm/tests/test_xrfi.py
```diff
@@ -26,22 +28,30 @@ def get_accuracy(f, rfi, verbose=VERBOSE):
...
+    SIZE = 100
```
adampbeardsley: Does this overwrite the global variable? Maybe this should be passed as an optional variable, with the default as the global SIZE.

hera_qm/tests/test_xrfi.py
```diff
@@ -54,45 +64,90 @@ def setUp(self):
...
+        mode = self.mode['detrend_deriv']
+        for i in range(3):
+            print args[i]
+            print cf[i],fp[i]
+            self._run_test(xrfi.detrend_deriv, args[i],  cf[i], fp[i], nsig=4, fmode=mode[i])
 
     def test_detrend_medfilt(self):
         cf, fp = self.ans['detrend_medfilt']
-        self._run_test(xrfi.detrend_medfilt, cf, fp, nsig=4)
+        argsList = [(8,8),(7,9),(9,7)]
```
adampbeardsley: Something for the future - we should think about how to get the argLists and modes and ans's more closely coupled in the code. Right now there's a bit of shifting back and forth to try to understand what is being run and why.

hera_qm/tests/test_xrfi.py
```diff
@@ -54,45 +64,90 @@ def setUp(self):
...
-        'detrend_medminfilt': (.97, .05),
-        'xrfi_simple': (.99, .1),
+        'detrend_deriv': [(.9, .6, .6),(.1,.005,.005)],
+        'detrend_medfilt': [(.9, .9, .9),(.01,.01,.01)],
+        'detrend_medminfilt': [(.97, .95, .95),(.5,.5,.5)],
+        'xrfi_simple': [(.99, .99),(.01, .01)],
         'xrfi': (.99, .01),
+        'watershed': [(.9,.9),(.1,.1)],
+    }
+    mode = {
```
adampbeardsley: Could you add a comment describing what `mode` means?

hera_qm/tests/test_xrfi.py
```diff
@@ -54,45 +64,90 @@ def setUp(self):
...
+        'detrend_deriv': [False,False,False],
+        'detrend_medfilt': [False,False,False],
+        'detrend_medminfilt': [False,False,False],
+        'xrfi_simple': [False,False],
+        'xrfi': False,
+        'watershed': [False,True],
```
adampbeardsley: Also, please add a comment explaining why you expect failure whenever you set mode to true (unless it's obvious from context).

hera_qm/xrfi.py
```diff
@@ -1,51 +1,116 @@
...
 def medmin(d):
-    """Calculate the median minus minimum statistic of array.
-
+    '''Calculate the median minus minimum statistic of array.
     Args:
-        d (array): 2D data array
-
+        d (array): 2D data array of the shape (time,frequency).
     Returns:
         (array): array with the statistic applied.
```
adampbeardsley: I don't think this returns an array. Looks like a single number to me.

hera_qm/xrfi.py
```diff
@@ -1,51 +1,116 @@
...
         for j in xrange(d.shape[1]):
-            i0, j0 = max(0, i - K), max(0, j - K)
-            i1, j1 = min(d.shape[0], i + K), min(d.shape[1], j + K)
+            i0, j0 = max(0, i - Kt), max(0, j - Kf)
+            i1, j1 = min(d.shape[0], i + Kt), min(d.shape[1], j + Kf)
             d_sm[i, j] = medmin(d[i0:i1, j0:j1])
     return d_sm
 
+def detrend_deriv(d, dt=True, df=True):
+    '''XXX This only works ok on sparse RFI.'''
```
adampbeardsley: Please add doc string with description, args, and returns. Here, and throughout, make important notes with: `NB: something something`... `XXX` is (regrettably) often used as a TODO in code, so let's avoid that confusion.

hera_qm/xrfi.py
```diff
@@ -196,7 +188,7 @@ def xrfi(d, f=None, K=8, sig_init=6, sig_adj=2):
     Returns:
         bool array: array of flags
     """
-    nsig = detrend_medfilt(d, K=K)
+    nsig = detrend_medfilt(d, Kt=Kt, Kf=Kf)
     f = watershed_flag(np.abs(nsig), f=f, sig_init=sig_init, sig_adj=sig_adj)
     return f
```
adampbeardsley: What is the trailing comment referring to? Was this a TODO long forgotten?

hera_qm/xrfi.py
```diff
@@ -1,51 +1,116 @@
...
+        Kf (int): size in frequency to apply medminfilter over
+    Returns:        
+         bool array: boolean array of flags    
+    """
+    d_sm = medminfilt(np.abs(d), 2 * Kt + 1, 2 * Kf + 1)
+    d_rs = d - d_sm
+    d_sq = np.abs(d_rs)**2
+    # puts minmed on same scale as average
+    if Kt == Kf:
+        sig = np.sqrt(medminfilt(d_sq, 2 * Kt + 1)) * (Kt / .64)
```
adampbeardsley: This isn't doing what you want - even if Kt == Kf, they may not equal 8, which is the default in medminfilt. I think you can safely use the `else` statement in all cases, so this if statement isn't really necessary.

hera_qm/xrfi.py
```diff
@@ -1,51 +1,116 @@
...
+    Returns:        
+        bool array: boolean array of flags    
+    """
+    d = np.concatenate([d[Kt - 1::-1], d, d[:-Kt - 1:-1]], axis=0)
+    d = np.concatenate([d[:, Kf - 1::-1], d, d[:, :-Kf - 1:-1]], axis=1)
+    d_sm = medfilt(d, kernel_size=(2 * Kt + 1, 2 * Kf + 1))
+    d_rs = d - d_sm
+    d_sq = np.abs(d_rs)**2
+    # puts median on same scale as average    
+    # .456 scaling?
```
adampbeardsley: Yeah, I don't know where this number comes from. @jsdillon is this similar to the scaling you use in a modified z-score?

hera_qm/tests/test_xrfi.py
```diff
@@ -4,12 +4,14 @@
...
 import hera_qm.tests as qmtest
+from inspect import getargspec
+
 
 np.random.seed(0)
 
 SIZE = 100
 VERBOSE = False
-PLOT = False
+LOT = False
```
jkerrigan: Yes, not sure how that happened.

hera_qm/tests/test_xrfi.py
jkerrigan: Will do.

hera_qm/tests/test_xrfi.py
jkerrigan: fake_flags is used in test_xrfi_simple (line 130) and test_watershed (line 145), this tests a failure mode(and coverage) of handing them an incorrect flag array and receiving an exceptionally low correct number of flags and high false positives.

hera_qm/tests/test_xrfi.py
```diff
@@ -26,22 +28,30 @@ def get_accuracy(f, rfi, verbose=VERBOSE):
...
+    RFI = 10
+    SIZE = 100
+    fakeflags = np.random.randint(0,2,size=(SIZE,SIZE)).astype(bool)
+    return fakeflags
+    
 
 def plot_waterfall(data, f, mx=10, drng=10, mode='lin'):
-    if not PLOT:
-        return
+    #if not PLOT:
```
jkerrigan: Fixed. I think it had to do with my capo installation.

hera_qm/xrfi.py
jkerrigan: I'm not worried about if they're equal to 8 here, I'm trying to make sure the scaling stays in line with the case that you're using a square kernel. The case where Kt!=Kf gives a scaling of sqrt(Kt**2 + Kf**2) which would have the square case be off by a factor of sqrt(2). Unless I'm misunderstanding what should be done here.

hera_qm/xrfi.py
jkerrigan: I'm not sure, there's no other reference to it besides this as far as I'm aware.

hera_qm/xrfi.py
jkerrigan: Fixed.

hera_qm/xrfi.py
adampbeardsley: Oh weird. We'd better try to understand the scaling factors. I wonder if the .64 is sqrt(2)*.456 (the scaling factor used elsewhere). But in general I wouldn't think a square would need a special case for the scaling.
Regardless of the scaling, the statement is still not getting the third argument to `medminfilt` passed in, so it's defaulting to `Kf=8`, which is not expected behavior.

adampbeardsley:
In testing the xrfi wrapper with Paul, we noticed a bug in xrfi. In detrend_medfilt (and possibly other functions) if the data size is smaller than the kernel size, the result is an empty array. We should make sure the result is always the appropriate size - this may mean adjusting the subscripting appropriately (possibly throwing a warning), and/or erroring.

PR #35, opened 2017-07-26
Merge firstcal metrics

nkern:
This adds the FirstCal_Metrics() class to hera_qm. This module operates on *.calfits files that are output by FirstCal, and uses metrics to determine whether the solution as a whole is "good" or "bad", and can identify individual antenna delay solutions that are "good" or "bad", depending on the user's tolerance. Plotting routines are included to visualize fluctuations on delay solutions.

Basic capabilities are displayed in the Jupyter Notebook demo in the scripts/ directory. Tests are included for each function in the FirstCal_Metrics() class.

Let me know if this checks out and if you need anything else.


plaplant:
@nkern your tests on Travis are failing right now, because it cannot connect to a display when you go to test the plot functions made with `matplotlib`. Though you've tried to do `matplotlib.use('Agg')` to change to a non-interactive backend in your test file, it's still somehow importing `matplotlib.pyplot` earlier. I'm not sure where exactly this is happening (my guess an `import aipy` is being executed before this), so you'd have to figure out where and put the 'Agg' statement before that.

More generally, the plotting routines are excellent, but I'm not sure the modules are the best place for them. Perhaps they're better off in the scripts folder, so that the user can invoke them from the command line. This would also side-step tracking down the errant matplotlib import statement to make the tests pass, since it looks like things are fine besides that. @adampbeardsley @jsdillon what do you think about moving the plotting functions to the scripts folder?

nkern:
Thanks @plaplant for the feedback. @jsdillon and @adampbeardsley let me know your thoughts when you get some time. 

About the plotting routines, its weird b/c the
```
import matplotlib
matplotlib.use('Agg')
```
is literally the first import in the test script. Regardless, I think we should be able to find a solution to the "DISPLAY not found" error. I'll investigate and see what I can do.

Regardless of whether they are attached to the class or not, I suspect that plotting routines will be used b/c I think it is important to be able to visualize the delay solutions. And, if they are going to be used, we should probably have unit tests for them. Which I think puts us back where we started.

SaulAryehKohn:
@nkern I solved a similar issue by including:
```
before_script:
    - "export DISPLAY=:99.0"
    - "sh -e /etc/init.d/xvfb start"
    - sleep 3
```
in my travis.yml file [here](https://github.com/UPennEoR/radionopy/blob/master/.travis.yml)

hera_qm/firstcal_metrics.py
```diff
@@ -0,0 +1,533 @@
...
+        # put into dictionary
+        result = OrderedDict()
+        result['full_sol'] = self.full_sol
+        result['bad_ant'] = self.bad_ants
+        result['z_scores'] = self.z_scores
+        result['ant_std'] = self.ant_std
+        result['time_std'] = self.time_std
+        result['agg_std'] = self.agg_std
+        result['times'] = self.times
+        self.result = result
```
adampbeardsley: We also want to include the code version in any files written. Have a look at ant_metrics for how to do this, and I can help if you need it.

hera_qm/firstcal_metrics.py
```diff
@@ -0,0 +1,533 @@
...
+
+        Uses astropy.stats.biweight_midvariance for a robust measure
+        of the std
+
+        Input:
+        ------
+
+        return_dict : bool, [default=False]
+            if True, return time_std, ant_std and z_scores
+            as a dictionary with "ant" as keys in str
```
adampbeardsley: Not essential for now, but we should try to be consistent between different modules - ant_metrics uses ants as integers as keys.

hera_qm/firstcal_metrics.py
```diff
@@ -0,0 +1,533 @@
...
+import cPickle as pkl
+
+
+class FirstCal_Metrics(object):
+    """
+    FirstCal_Metrics class for holding firstcal data,
+    running metrics, and plotting delay solutions
+
+    FirstCal_Metrics currently only works on a 
+    per-file basis, as pyuvadata.UVCal also only
```
nkern: great, sounds good

hera_qm/firstcal_metrics.py
```diff
@@ -0,0 +1,533 @@
...
+
+        self.delsys : ndarray, shape=(N_ant, N_times)
+            firstcal delay solutions
+
+        self.delay_offsets : ndarray, shape=(N_ant, N_times)
+            firstcal delay solution offsets from time average
+
+        self.frac_JD : ndarray, shape=(N_times,)
+            ndarray containing time-stamps of each integration
+            in fraction of current JD
```
nkern: `frac_JD` means remainder of JD after subtracting the integer value. I'm open to whatever is the clearest, so let me know if you have any ideas

hera_qm/firstcal_metrics.py
nkern: sounds good

hera_qm/firstcal_metrics.py
```diff
@@ -0,0 +1,533 @@
...
+        result['ant_std'] = self.ant_std
+        result['time_std'] = self.time_std
+        result['agg_std'] = self.agg_std
+        result['times'] = self.times
+        self.result = result
+
+        if output == True:
+            return self.result
+       
+    def write_metrics(self, filename=None, filetype='pkl'):
```
nkern: One thing I'm finding is that you can't JSON-ify an ndarray, which I'm not a big fan of b/c I like ndarrays, and I prefer not to have to make an ndarray a list to write it out, and then convert back to an ndarray when we load it in. However, we can default to a JSON and I'll keep the pickle functionality included.

hera_qm/firstcal_metrics.py
```diff
@@ -0,0 +1,533 @@
...
+            if None, use default filename stem
+
+        filetype : str, default='pkl'
+            filetype to write out to
+            options = ['json', 'pkl']
+
+        """
+        # get filename prefix
+        if filename is None:
+            filename = self.file_stem + ".metrics"
```
nkern: `file_stem` has a *.first.* in it already that comes from the FirstCal solution.

hera_qm/firstcal_metrics.py
nkern: Also, to load in a JSON and turn it into a dictionary it seems like we need to use the ```eval``` function which is something I try to never do. Pickle just seems a whole lot easier to deal with if we are going to be writing data structures, like dictionaries and ndarrays.

nkern: nvm, found a way to JSON-ify w/o ```eval```

adampbeardsley:
I just pushed a couple changes:
1) More test coverage. The module is now at 97% covered -- feel free to finish it off to make it 100% :)
2) I added the average delay to the output files. @nkern Please take a look at that code. Note this changes the output of delay_std(), and I haven't checked the ipython notebook.

Because I made these changes, it would be good to have another set of eyes before merging. @SaulAryehKohn, @jsdillon, or @plaplant would any of you be willing to finish off this review?

plaplant:
@adampbeardsley sure I can take a look at it. One issue I had was that tests were not passing locally for me before this push, so I'll pull the changes and see if they pass now

nkern:
Thanks @adampbeardsley for helping me out w/ all the tests! I'll try and add more to get it up further. I'll take a closer look at your additions to see if I can spot any bugs.

hera_qm/firstcal_metrics.py
```diff
@@ -0,0 +1,544 @@
...
+
+            # set colorbar
+            fig.colorbar(cax, label='z-score')
+
+        if plot_type == 'time_avg':
+            # plot
+            cmap = matplotlib.cm.spectral(np.linspace(0, 0.95, len(self.ants)))
+            z_scores = np.mean(z_scores, axis=1)
+            ax.grid(True)
+#            ax.scatter(range(len(z_scores)), z_scores,  c=cmap, alpha=0.85,
```
plaplant: Same comment as above re: commented out vs. removed

hera_qm/firstcal_metrics.py
```diff
@@ -0,0 +1,544 @@
...
+                          np.floor(self.frac_JD.max() * 1000) / 1000]
+            ylen = [0, len(self.ants)]
+            cax = ax.matshow(z_scores, origin='lower', aspect='auto',
+                             extent=[xlen[0], xlen[1], ylen[0], ylen[1]], **kwargs)
+
+            # define ticks
+            ax.xaxis.set_ticks_position('bottom')
+            xticks = np.arange(xlen_round[0], xlen_round[1] + 1e-5, 0.001)
+            ax.set_xticks(xticks)
+            # [t.set_rotation(20) for t in ax.get_xticklabels()]
```
plaplant: If this line is not needed, it should be removed instead of being commented out.

hera_qm/firstcal_metrics.py
```diff
@@ -0,0 +1,544 @@
...
+
+        """
+        # calculate standard deviations
+        ant_avg = self.delay_avgs
+        ant_std = np.sqrt( astats.biweight_midvariance(self.delay_offsets, axis=1) )
+        time_std = np.sqrt( astats.biweight_midvariance(self.delay_offsets, axis=0) )
+        agg_std = np.sqrt( astats.biweight_midvariance(self.delay_offsets) )
+
+        # calculate z-scores
+        z_scores = np.abs(self.delay_offsets / agg_std)
```
plaplant: Why are we taking the absolute value? It might be better to leave the sign information in, and then take the absolute value when you go to plot or make a cut. But in principle I think this throws away information that could be useful.

.travis.yml
```diff
@@ -30,6 +31,11 @@ install:
   - pip install git+https://github.com/HERA-Team/linsolve.git
   - python setup.py install
 
+before_script:
```
plaplant: Great job getting this working with Travis.

setup.py
```diff
@@ -38,6 +38,7 @@ def package_files(package_dir, subdirectory):
     'version': version.version,
     'package_data': {'hera_qm': data_files},
     #    'install_requires': ['numpy>=1.10', 'scipy', 'pyuvdata', 'astropy>1.2', 'aipy']
```
plaplant: This line (and the 'dependency_links' one) should be deleted if not needed

.travis.yml
nkern: thank @SaulAryehKohn for that!

hera_qm/firstcal_metrics.py
```diff
@@ -0,0 +1,542 @@
...
+        metrics['ant_avg'] = self.ant_avg
+        metrics['ant_std'] = self.ant_std
+        metrics['time_std'] = self.time_std
+        metrics['agg_std'] = self.agg_std
+        metrics['times'] = self.times
+        metrics['version'] = self.version_str
+        self.metrics = metrics
+
+        if output is True:
+            return self.metrics
```
nkern: 1) Yup, I just updated it

2) Yeah, I guess its just for convenience.. I don't feel strongly about it either way.

hera_qm/firstcal_metrics.py
nkern: yeah sure

setup.py
nkern: I'll leave that to @adampbeardsley b/c I haven't been editing the setup.py script

nkern:
For some reason I'm now getting the following error on folio when I try to load any modules that use matplotlib.pyplot
```
QXcbConnection: Could not connect to display 
Aborted (core dumped)
```
not sure why this is happening, anyone else getting this? This just started a few hours ago for me. In anycase, a fix I found [online](https://github.com/xdata-skylark/libskylark/issues/63) is
```
export QT_QPA_PLATFORM=offscreen
```
which works for me. Not sure how or why this started happening for me

hera_qm/firstcal_metrics.py
```diff
@@ -0,0 +1,533 @@
...
+
+        # Instantiate UVCal and read calfits
+        self.UVC = UVCal()
+        self.UVC.read_calfits(calfits_file)
+
+        # get file prefix
+        self.file_stem = '.'.join(calfits_file.split('.')[:-1])
+
+        # Calculate median delay
+        self.delays = self.UVC.delay_array.squeeze() * 1e9
```
nkern: resolved

hera_qm/firstcal_metrics.py
```diff
@@ -0,0 +1,533 @@
...
+        self.times = self.UVC.time_array
+        self.start_JD = np.floor(self.times).min()
+        self.frac_JD = self.times - self.start_JD
+        self.minutes = 24 * 60 * (self.frac_JD - self.frac_JD.min())
+        self.Nants = self.UVC.Nants_data
+        self.ants = self.UVC.ant_array
+
+    def run_metrics(self, std_cut=0.5, output=False):
+        """
+        Run all metrics and write results to file
```
nkern: cleared up

hera_qm/firstcal_metrics.py
nkern: updated

hera_qm/firstcal_metrics.py
nkern: good point

hera_qm/firstcal_metrics.py
nkern: yup

hera_qm/firstcal_metrics.py
plaplant: Alright, then I think that my recommendation is to get rid of this. I think it's a little confusing anyway for a function to return a value only if asked to, and the name (`run_metrics`) implies that it just computes the metrics (as opposed to, say `get_metrics`). But I think the name should stay as is, since it's nice and straightforward.

hera_qm/firstcal_metrics.py
```diff
@@ -0,0 +1,579 @@
...
+
+        custom_ax = True
+        if ax is None:
+            custom_ax = False
+            fig = plt.figure(figsize=figsize)
+            ax = fig.add_subplot(111)
+
+        # Get zscores
+        ant_avg, ant_std, time_std, agg_std, z_scores = self.delay_std()
+        z_scores = np.abs(z_scores)
```
plaplant: Now that you're saving the sign information of the z-scores (which is great!), maybe you could have plotting the absolute value be an option of the routine (perhaps the default), and an option to plot the signed z-score. Just make sure that in the latter case, you switch to a diverging color map (like `spectral` in your previous routine) instead of `viridis_r`. (Side note: in my book, you get +10 points for using `viridis` instead of `jet` ;) )

setup.py
plaplant: Okay, I'll make a quick change to take this out.

adampbeardsley:
I just noticed an issue we're going to have when putting these metrics into the db - we need a polarization associated with these metrics. I don't think it will be too hard to add, but it will mean changing the keys to the metrics dictionary. I don't want to slow down this PR, so I'm happy to start a new branch once this gets merged.

adampbeardsley:
Two additional things that will help downstream processing:
- Could we store a per-antenna z_score in the metrics dict (averaging over time dimension in some fashion)?
- It would be helpful to have `full_sol` indicated with a boolean (rather than 'good' and 'bad'). Alternatively we could have an additional variable which repeats the info as a boolean so we don't change the current setup.

Right now I'm handling these things as special cases in the script to put metrics in M&C. It would really clean up the script if I could get rid of these special cases.

hera_qm/firstcal_metrics.py
nkern: Sure, I can add that. Making the function change to a diverging colormap automatically if plot_abs == True is a little tricky though: I currently allow the user to feed plot_kwargs, so I think it might be best to leave it up to the user to change colormaps if they want.

nkern:
@adampbeardsley sure can do both. changed it to "good_sol" is either True or False to make things a little more clear. Also added the key "ant_z_score" which is the z_scores ndarray averaged over time

hera_qm/firstcal_metrics.py
plaplant: Oh good point, and I see that `vmin` and `vmax` (which would also have to change) are also kwargs. Okay, then we'll trust in the user to do the sensible thing (famous last words...)

hera_qm/firstcal_metrics.py
```diff
@@ -0,0 +1,579 @@
+"""
+FirstCal metrics
+"""
+import matplotlib
+import matplotlib.pyplot as plt
+plt.switch_backend('Agg')
```
adampbeardsley: Is this line still necessary with the travis fix? It's messing up my environment after I import anything from hera_qm.

nkern: ah, perhaps not. I added that the same time I added the line to the travis file. I'll take it out and see if travis still works.

nkern:
In this most recent commit, I changed the structure of the FirstCal_Metrics class a little bit. The biggest thing I did was pull the `plot_std()` and `plot_zscores()` functions out of the class. The major reason I did this was so that if you want to inspect a *.first_metrics.json file, you can plot the metrics without having to instantiate a FirstCal_Metrics class and load a FirstCal calfits file.

I also added some more tests.

nkern:
I think I'm ready to merge this if people are happy with how it stands.

nkern:
*and it if passes the build...

hera_qm/firstcal_metrics.py
```diff
@@ -0,0 +1,542 @@
...
+
+        """
+        # get filetype
+        filetype = filename.split('.')[-1]
+        if filetype == 'json':
+            with open(filename, 'r') as f:
+                self.metrics = json.load(f)
+
+            # ensure keys of dicts are not strings
+            for i in self.metrics['ant_avg'].keys():
```
adampbeardsley: I spent too much time yesterday trying to get this to work. ant_metrics uses tuples as keys, which json doesn't like. As far as I can tell, the best thing to do is convert them to strings, but then eval is necessary. All this to say we should revisit when we're unifying the two classes.

Issue #36, opened 2017-07-28
Unify metric structures


adampbeardsley:
This will be aided by addressing issue #31, but we need to work on keeping code/calculations uniform across the package. For example, right now ant_metrics uses integers for antenna keys into dictionaries, while firstcal_metrics uses strings. We should also look into uses the same calculations - modified z scores vs zscores using biweight_midvariance. The metric base class should help a lot with this.

adampbeardsley:
This should also involve revisiting the interface with hera_mc. The `get_X_metrics_dict` functions, and the `metrics2mc` functions should be somewhat streamlined, and broad documentation is needed to instruct developers how to add/change metrics we use.

PR #47, opened 2017-08-07
QM to M&C interface

adampbeardsley:
This branch provides a function to convert metrics files to a simple dictionary that M&C can use to populate the db. By introducing this function M&C doesn't need to "know" about the internal structures we have for various metrics, and we can change things without breaking M&C (as long as we update the metrics2mc function.
I also added dictionaries of metrics and descriptions to the top of each metric module (ant_metrics, firstcal_metrics, and the new omnical_metrics). These are used by M&C to populate a table of metrics with their descriptions. Any change to what metrics we calculate/store should be updated here as well.


hera_qm/utils.py
```diff
@@ -138,3 +140,79 @@ def generate_fullpol_file_list(files, pol_list):
                 file_list.append(jd_list)
 
     return file_list
+
+
+def metrics2mc(filename, ftype):
+    """ Read in file containing quality metrics and stuff into a dictionary which
```
plaplant: Could we document how to add a new value to the dictionary we pass to M&C? It's not totally clear to me if it's enough to add it to the metrics dictionary or whether the module-level dictionaries need to be edited as well. Perhaps we can add it to the README? Or write a memo?

hera_qm/ant_metrics.py
```diff
@@ -10,6 +10,32 @@
 from hera_qm import utils
 
 
+ant_metrics_list = {'ant_metrics_meanVij': 'Mean of the absolute value of all '
```
adampbeardsley: Yes, I will add comments explaining these. Actually, I wonder if it would be cleaner to make a function, `get_metrics_dict` or something similar, which just returns this dict. That way I can add a docstring, and it will be a little more clear when things are coming from. Would that make more sense?

hera_qm/utils.py
```diff
@@ -138,3 +140,79 @@ def generate_fullpol_file_list(files, pol_list):
...
+            d['ant_metrics']['ant_metrics_removal_iteration'] = []
+            metric = 'ant_metrics_removal_iteration'
+            for antpol, val in data['removal_iteration'].items():
+                d['ant_metrics'][metric].append([antpol[0], antpol[1], val])
+
+    elif ftype is 'firstcal':
+        from hera_qm.firstcal_metrics import load_firstcal_metrics
+        data = load_firstcal_metrics(filename)
+        pol = data['pol']
+        d['array_metrics']['firstcal_metrics_good_sol'] = float(data['good_sol'])
```
adampbeardsley: Actually, this gets recast in M&C now, so the float part here isn't necessary. I will take it out.

hera_qm/ant_metrics.py
plaplant: Yeah, I think putting them in a function is perfect. I also think having some broader documentation is helpful, but it would probably be too involved for a docstring. We should add the task of writing it to the issues page so it doesn't get lost in the shuffle.

adampbeardsley: Ok. I'm working on the functions, will push a commit soon. I also added a comment to the existing issue #36 (unify metric structures) to make sure we revisit the broad documentation. I'm reluctant to write up a memo right now because I _hope_ the interface gets a lot more simple when we unify things.

hera_qm/utils.py
adampbeardsley: In the commit that's on its way I've added to the docstring explaining how to add a metric. But the long-term solution will be to add a memo and link in the readme once we undergo other reorganizations (a la issue #36).

PR #54, opened 2017-08-18
Untangle dependencies

adampbeardsley:
This PR takes the datacontainer out of hera_qm (it has already been moved to hera_cal). I also reordered some of the installs to make things work, and added tests to ant_metrics to keep the total coverage above 95%. I also updated data files to work with a pyuvdata update and not spit out a ton of warnings.


hera_qm/vis_metrics.py
```diff
@@ -1,35 +1,27 @@
...
 
     Args:
-        data (dict): dictionary of visibilities with keywords of pol/ant pair
-            in any order.
-        wgts (dict): dictionary of weights with keywords of pol/ant pair in any order
-        bandwidth (float): Channel width, preferably Hz, but only restriction is
-            bandwidth units are recipricol inttime units.
-        inttime (float): Integration time, preferably seconds, but only restriction
-            is inttime units are recipricol bandwidth units.
+        data (UVData): UVData object with data.
```
plaplant: This is a great change, it's much more straightforward to be using a UVData object

hera_qm/tests/test_ant_metrics.py
```diff
@@ -37,26 +37,50 @@ def test_mean_Vij_metrics(self):
                                                 self.ants, self.bls, rawMetric=True)
         self.assertEqual(mean_Vij, {(1, 'y'): 1.0, (1, 'x'): 1.0, (0, 'x'): 1.0,
                                     (0, 'y'): 1.0, (2, 'x'): 1.0, (2, 'y'): 1.0})
+        args = [self.data, self.pols, self.antpols, self.ants, self.bls]
+        zs = uvtest.checkWarnings(ant_metrics.mean_Vij_metrics, args, nwarnings=6,
```
adampbeardsley: Yeah, they're divide by zero. The problem is that the fake data is all ones, and calculating the modified-z for constant data gives nans. I thought about making the data more interesting, but then we would need to figure out what the expected result is to check it. I can change it if you don't think this is a valid test.

.travis.yml
```diff
@@ -26,10 +26,10 @@ install:
   - conda install -c conda-forge aipy
   - pip install coveralls
   - pip install git+https://github.com/HERA-Team/pyuvdata.git
-  - pip install git+https://github.com/HERA-Team/hera_cal.git
+  - python setup.py install  # Needs to be installed before hera_cal
```
adampbeardsley: I agree this is a pain. It's not due to datacontainer though, it's because hera_cal uses ant_metrics. I think it just uses it to read in the metrics to flag bad ants, so we could potentially move that import to within a function, then we can put these back in order.

hera_qm/tests/test_vis_metrics.py
```diff
@@ -2,42 +2,24 @@
...
-        for bl in data['xx']:
-            n = nos[bl + ('xx',)]
-            self.assertEqual(n.shape, (nchan - 1,))
-            np.testing.assert_almost_equal(n, 10 * np.ones_like(n), -1)
-        nos = vis_metrics.check_noise_variance(data, wgts, 10., 10.)
-        for bl in data['xx']:
-            n = nos[bl + ('xx',)]
-            self.assertEqual(n.shape, (nchan - 1,))
-            np.testing.assert_almost_equal(n, 100 * np.ones_like(n), -2)
+        nos = vis_metrics.check_noise_variance(self.data)
```
adampbeardsley: This was exactly my reasoning. I'll look into a more controlled UVData object.

.travis.yml
plaplant: Yes, I like the idea of burying the import in hera_cal until we need it. I can make that change (either before or after we finish this PR)

hera_qm/tests/test_ant_metrics.py
plaplant: Oh I see. In that case, I think we should probably change it. My biggest concern is that these warnings might be hiding some more sinister bugs, so if we could eliminate those warnings that would be great. Maybe we can have simple-but-not-uniform data, something like an even distribution of 1s, 2s, and 3s? So our median would be 2, and the modified z-score would be calculable. (Haven't thought through all the specifics, but I can cook up some simple data with a known result if you'd like me to.)

adampbeardsley:
Ok @plaplant - I think I've addressed your questions/suggestions in the latest series of commits. Let me know if you see anything else that needs fixing.

hera_qm/tests/test_ant_metrics.py
```diff
@@ -26,37 +27,71 @@ class TestLowLevelFunctions(unittest.TestCase):
...
         self.antpols = ['x', 'y']
-        self.bls = [(0, 1), (1, 2), (0, 2)]
+        self.bls = [(0, 1), (1, 2), (2, 3), (0, 2), (1, 3), (0, 3)]
 
     def test_mean_Vij_metrics(self):
         mean_Vij = ant_metrics.mean_Vij_metrics(self.data, self.pols, self.antpols,
                                                 self.ants, self.bls, rawMetric=True)
-        self.assertEqual(mean_Vij, {(1, 'y'): 1.0, (1, 'x'): 1.0, (0, 'x'): 1.0,
-                                    (0, 'y'): 1.0, (2, 'x'): 1.0, (2, 'y'): 1.0})
+        # The reference dictionaries here and in other functions were determined
```
plaplant: Excellent change, and the documentation is very helpful.

hera_qm/tests/test_vis_metrics.py
```diff
@@ -2,42 +2,45 @@
...
+import pyuvdata.tests as uvtest
+import copy
 
 
 class TestMethods(unittest.TestCase):
 
     def setUp(self):
-        self.data1 = qmtest.noise(size=(100, 100))
-        self.data2 = qmtest.noise(size=(100, 100))
+        self.data = UVData()
```
plaplant: Fantastic work getting things to be similar, thanks!

Issue #53, opened 2017-08-18
Summarize rfi


adampbeardsley:
We need a function/script to summarize the rfi flagging on an observation. This should somehow average over baselines/antennas/times/whathaveyou to result in a spectrum of maximum flagging, minimum flagging, mean, and std. This information needs to go in the library and/or M&C.

adampbeardsley:
Most of this is addressed with the above reference merge. But I'm going to leave the issue open because there is still a desire to have a nightly summary. This is something we will likely address once we have a stable H1C observing season under way.

adampbeardsley:
@nkern could you get the rfi summary info into your notebooks? In particular, we're generating flag summary npz's with useful info.

adampbeardsley:
Nick added a couple plots to the notebook. Which made us realize there is a bug where the summary is being computed on the prior flags, rather than the new ones. I'm working on a fix, will submit PR soon, and we can hopefully get it into the "frozen" code.

adampbeardsley:
Don't know exactly when, but this has been fixed for a while.

PR #58, opened 2017-08-22
Summarize rfi

adampbeardsley:
Create npz file with synthesized rfi info. Addresses Issue #53.

### DIFF
--- a/hera_qm/xrfi.py
+++ b/hera_qm/xrfi.py
@@ -315,7 +315,25 @@ def summarize_flags(uv, outfile):
         outfile -- filename for output npz file
 
     Return:
-        npz file containing "spectrum" of max, min, mean, std, median of flags
+        Writes an npz file to disk containing summary info. The keys are:
+            waterfall - ndarray (Ntimes, Nfreqs, Npols) of rfi flags, averaged
+                        over all baselines in the file.
+            tmax - ndarray (Nfreqs, Npols) of max rfi fraction along time dimension.
+            tmin - ndarray (Nfreqs, Npols) of min rfi fraction along time dimension.
+            tmean - ndarray (Nfreqs, Npols) of average rfi fraction along time dimension.
+            tstd - ndarray (Nfreqs, Npols) of standard deviation rfi fraction
+                   along time dimension.
+            tmedian - ndarray (Nfreqs, Npols) of median rfi fraction along time dimension.
+            fmax - ndarray (Ntimes, Npols) of max rfi fraction along freq dimension.
+            fmin - ndarray (Ntimes, Npols) of min rfi fraction along freq dimension.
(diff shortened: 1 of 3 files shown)


hera_qm/xrfi.py
```diff
@@ -299,4 +301,51 @@ def xrfi_run(files, opts, history):
...
+
+
+def summarize_flags(uv, outfile):
+    """ Collapse several dimensions of a UVData flag array to summarize.
+    Args:
+        uv -- UVData object containing flag_array to be summarized
+        outfile -- filename for output npz file
+
+    Return:
+        npz file containing "spectrum" of max, min, mean, std, median of flags
```
plaplant: Can we add some documentation of what's in these files? Maybe we should explicitly state what keys are present and a quick description. It would be helpful for reading them back in later.

hera_qm/xrfi.py
```diff
@@ -299,4 +301,51 @@ def xrfi_run(files, opts, history):
...
+    fmax = np.max(waterfall, axis=1)
+    fmin = np.min(waterfall, axis=1)
+    fmean = np.mean(waterfall, axis=1)  # no weights - all bls have all channels
+    fstd = np.std(waterfall, axis=1)
+    fmedian = np.median(waterfall, axis=1)
+    # Some meta info
+    freqs = uv.freq_array[0, :]
+    pols = uvutils.polnum2str(uv.polarization_array)
+    hera_qm_version = hera_qm_version_str
+    # Store data in json
```
plaplant: store data in npz?

hera_qm/xrfi.py
```diff
@@ -299,4 +301,51 @@ def xrfi_run(files, opts, history):
...
+    # Calculate stats across frequency
+    fmax = np.max(waterfall, axis=1)
+    fmin = np.min(waterfall, axis=1)
+    fmean = np.mean(waterfall, axis=1)  # no weights - all bls have all channels
+    fstd = np.std(waterfall, axis=1)
+    fmedian = np.median(waterfall, axis=1)
+    # Some meta info
+    freqs = uv.freq_array[0, :]
+    pols = uvutils.polnum2str(uv.polarization_array)
+    hera_qm_version = hera_qm_version_str
```
plaplant: Could we also save the version hash? If this info is already in the version string, then that's fine. (Mostly interested in being able to recreate the analysis that was run.)

hera_qm/xrfi.py
adampbeardsley: woops, will fix.

hera_qm/xrfi.py
adampbeardsley: Yup, that's already in there. That string has the "version" (currently 0.0.0), git origin, hash, branch, and description.

hera_qm/xrfi.py
adampbeardsley: Will do.

PR #59, opened 2017-08-24
Swapped out scipy medfilt functions for speed improvements

jkerrigan:
As far as I can tell for our purpose of using the medfilt for waterfalls we can switch to medfilt2d which for some reason has different scipy.signal dependencies and that allows it to run much quicker. I've tested both medfilts on several datasets and they appear to be identical.

### DIFF
--- a/hera_qm/xrfi.py
+++ b/hera_qm/xrfi.py
@@ -108,17 +108,17 @@ def detrend_medfilt(d, Kt=8, Kf=8):
         bool array: boolean array of flags
     """
     # Delay import so scipy is not required for any use of hera_qm
-    from scipy.signal import medfilt
+    from scipy.signal import medfilt2d
 
     if Kt > d.shape[0] or Kf > d.shape[1]:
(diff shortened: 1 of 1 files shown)


adampbeardsley:
What do you mean that medfilt2d "has different scipy.signal dependencies"? Is there a specific version of scipy we need to require now?

jkerrigan:
Sorry for the confusion I worded that poorly, it's the same version of scipy but medfilt2d appears to use a different scipy script to do the 2d medfiltering. They even state here in this init script that its faster. 
https://github.com/scipy/scipy/blob/93bf1a6a494fef15b383831529543c7c17635f24/scipy/signal/__init__.py .

adampbeardsley:
I just merged this. @jkerrigan Did you want to keep the branch for further development, or should I go ahead and delete it?

jkerrigan:
I have a potential new xrfi technique I'd like to add, but I can do a separate branch for that if you want.

PR #63, opened 2017-09-06
Find off ants

jsdillon:
Here's a couple of quick fixes that help the problem when much of the array is recording zeros. This update to ant_metrics will find antennas who are involved in mostly zero visibilities and "pre-remove" them. They don't get metrics or z-scores, but they do appear as dead antennas, in xants, and have removal_iteration of -1.

I'm also changing the averaging of the redcorr and meanVij-based metrics to a `np.nanmean()` so that if one metric works but the other is a NaN, we can still get a result and act on it.

They are  @plaplant I leave it up to you how you want to integrate this what you're doing.


adampbeardsley:
@plaplant @jsdillon What's the status of this branch/PR? Does it just need unittests?

jsdillon:
Yes. I am writing a unittest.

jsdillon:
Ok, I'm ready for a review, @adampbeardsley or @plaplant or both of you.

This PR address issue #71.

jsdillon:
@adampbeardsley It was developed based on some of the files Paul had me looking at. It still breaks when more than half of the antennas are off (since then the median antenna is off), but that feels like a pathological case to me. 

I was looking at 2458000.46582 and 2458001.60112 when I wrote this. I forget which had which problems.

plaplant:
@adampbeardsley I believe the `2458000.*` files were the pathological ones, with almost all of the amplifiers off. I'm honestly not worried about that edge case, since there's little meaningful data if that's the case. As long as we're not getting `nan` across the array when a few antennas are down (the previous behavior), I'm happy with the fixes.

My only question is that now that we've split out `compute_median_auto_power_dict` to a separate function (which I think is the right move), do we want to design a specific test for it? If not, then we can merge this as is.

jsdillon:
My inclination is that it's not necessary. I only split it off because I went from using it once to using it twice. Also, it's such a simple function and it's hard to imagine how it should be tested in a way that's not brittle if we decide to change algorithms later.

plaplant:
Okay that sounds fair enough.

adampbeardsley:
I'm late to the party, but I was writing a test. I'll push and let @plaplant decide if it's worthwhile or not.

jsdillon:
Ok by me.

plaplant:
Thanks @adampbeardsley! Extra tests are always appreciated 😄

PR #64, opened 2017-09-06
Depricate optparse

adampbeardsley:
This is not at all urgent, but we should get away from optparse.


hera_qm/utils.py
```diff
@@ -244,8 +249,7 @@ def metrics2mc(filename, ftype):
             except KeyError:
                 raise ValueError('Invalid polarization for ant_metrics in M&C.')
             for ai, ant in enumerate(uvcal.ant_array):
-                val = np.median(uvcal.quality_array[ai, 0, :, :, pi], axis=0)
-                val = np.mean(val)
+                val = np.mean(uvcal.quality_array[ai, 0, 0, :, pi])
```
plaplant: Why have we gone from taking a median in frequency, then the mean in time, to just the mean in time of the first frequency? If we're only going to pick a single frequency, perhaps it should be something closer to the center of the band, since the edges are known to have poor solutions and (at the low-end of the band) heavy RFI contamination.

adampbeardsley: Err, I think this was when I misunderstood how chisq was being calculated to begin with. I vaguely remember thinking that it's the same for all frequencies, so it didn't really matter. But that's not the case, and it shouldn't go in the PR anyway. I've reverted it. Good catch!

PR #76, opened 2017-09-18
Broadcast flags

adampbeardsley:
Added functionality to broadcast flags across data based on thresholds. This is currently labeled "DO NOT MERGE" because it will affect the RTP workflow, but there are some defaults we need to decide on (namely, the thresholds).

### DIFF
--- a/hera_qm/tests/test_xrfi.py
+++ b/hera_qm/tests/test_xrfi.py
@@ -464,5 +464,21 @@ class TestSummary(unittest.TestCase):
         self.assertEqual(data['version'], hera_qm_version_str)
 
 
+class TestBroadcast(unittest.TestCase):
+    def test_summarize_flags(self):
+        from pyuvdata import UVData
+
+        infile = os.path.join(DATA_PATH, 'zen.2457698.40355.xx.HH.uvcAA')
+        uv = UVData()
+        uv.read_miriad(infile)
+        uv.flag_array[0, 0, uv.Nfreqs / 2, 0] = True
+        bflags = xrfi.broadcast_flags(uv, threshold=0.)
+        nbl = np.sum(uv.time_array == uv.time_array[0])
+        self.assertEqual(bflags.mean(), float(nbl) / (uv.Nblts * uv.Nfreqs))
+        # Check thresholding works correctly
+        bflags = xrfi.broadcast_flags(uv, threshold=0.5)
+        self.assertEqual(bflags.sum(), 1)
+
+
 if __name__ == '__main__':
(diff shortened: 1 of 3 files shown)


hera_qm/tests/test_xrfi.py
```diff
@@ -464,5 +464,35 @@ def test_summarize_flags(self):
...
+        bflags = xrfi.broadcast_flags(uv, bl_threshold=0.)
+        nbl = np.sum(uv.time_array == uv.time_array[0])
+        self.assertEqual(bflags.sum(), nbl)
+        # Check thresholding works correctly
+        bflags = xrfi.broadcast_flags(uv, bl_threshold=0.5)
+        self.assertEqual(bflags.sum(), 1)
+
+        # Test frequency thresholding
+        t_ind = np.where(uv.time_array == uv.time_array[0])[0]
+        uv.flag_array[t_ind, 0, 0:(uv.Nfreqs * 3 / 4), 0] = True
```
plaplant: To future-proof ourselves, we could add `from __future__ import division` at the top of the script, and change this to invoke explicit integer division: `0:(uv.Nfreqs * 3 // 4)`. Or we could cast the result to `int`, either way.

hera_qm/xrfi.py
```diff
@@ -220,13 +220,22 @@ def xrfi(d, f=None, Kt=8, Kf=8, sig_init=6, sig_adj=2):
...
-def xrfi_run(files, args, history):
+def xrfi_run(files, args, history, broadcast=True, bl_threshold=0., freq_threshold=0.9,
+             time_threshold=0.9):
     """
     Run an RFI-flagging algorithm on an entire file and store results in flag array.
 
     Args:
        files -- a list of files to run RFI flagging on
        args -- parsed arguments via argparse.ArgumentParser.parse_args
+       broadcast -- (bool) If True (default), broadcast flags across data if
```
plaplant: Can we add the broadcast bool and threshold levels as options for the ArgumentParser instead? It's a little messy to mix command line arguments and keyword args.

adampbeardsley: Whoops, I had that change sitting locally and forgot to commit. It's on its way.

hera_qm/tests/test_xrfi.py
adampbeardsley: Fixed.

adampbeardsley:
I've removed the "DO NOT MERGE" tag - originally I had broadcast set to true by default, but now it's defaulted to false, so it shouldn't kill RTP if we merge.

PR #80, opened 2017-09-21
Omnical metrics

nkern:
This PR adds `omnical_metrics.py` and relevant tests, CLI script and example ipython notebook.

One outstanding thing that still needs to be addressed is `get_omnical_metrics_dict()` in the first few lines of the `omnical_metrics.py` file. Perhaps @adampbeardsley you can tell me how you want this configured after you've had a glance at the code. It follows a similar format to `firstcal_metrics.py`


nkern:
currently working on adding more tests to increase coverage...

hera_qm/omnical_metrics.py
```diff
@@ -12,3 +28,796 @@ def get_omnical_metrics_dict():
...
+
+        phs_noise_good_sol : bool
+            boolean with the metric's good / bad determination for
+            the entire solution.
+        """
+        # take rfft
+        rfft = medfilt(np.abs(np.fft.rfft(phs_diff[:, :, self.band, :], axis=2)), kernel_size=(1, 1, 15, 1))
+
+        # Get phase noise
+        phs_noise = np.median(rfft[:, :, -100:, :], axis=2)
```
adampbeardsley: I'm not a big fan of hardcoded numbers, here and the `15` above. Could these be programmatically determined based on the number of channels and/or bandwidth? Or at least optional input?

hera_qm/omnical_metrics.py
```diff
@@ -12,3 +28,796 @@ def get_omnical_metrics_dict():
...
+        """
+        if hasattr(self, 'metrics') == False:
+            raise Exception("Must run self.run_metrics() before plotting routines...")
+
+        if self.firstcal_file is None:
+            raise Exception("Must supply firstcal_file in order to plot phase metrics...")
+
+        plot_phs_metric(self.metrics, plot_type=plot_type, ax=ax, save=save, fname=fname, outpath=outpath, **kwargs)
+
+    def plot_gains(self, ants=None, time_index=0, jones_index=0, divide_fc=False, plot_type='phs', ax=None,
```
adampbeardsley: Why are some of the plotting routines functions outside the class, with simple wrappers in the class, while others (like this one) only live in the class?

hera_qm/omnical_metrics.py
```diff
@@ -12,3 +28,796 @@ def get_omnical_metrics_dict():
...
+        # assign fc filename
+        self.firstcal_file = firstcal_file
+
+        # select freq channels
+        if cut_band == True:
+            self.band = np.arange(self.Nfreqs/10, self.Nfreqs - self.Nfreqs/10)
+        else:
+            self.band = np.arange(self.Nfreqs)
+
+        # Get robust standard deviation of chisq for each antenna
```
adampbeardsley: Even though it's brief, you may want to break these calculations into a separate function like you do for the phase metrics. That will be more consistent, and making testing a lot easier.

hera_qm/omnical_metrics.py
```diff
@@ -12,3 +28,796 @@ def get_omnical_metrics_dict():
...
+        metrics['ant_array']           = self.ant_array
+        metrics['jones']               = self.jones
+        metrics['pols']                = self.pols
+        metrics['Npols']               = self.Npols
+        metrics['times']               = self.times
+        metrics['Ntimes']              = self.Ntimes
+        metrics['Nants']               = self.Nants
+
+        metrics['version']             = self.version_str
+        metrics['history']             = self.history
```
adampbeardsley: As far as I can tell, this is always an empty string.

adampbeardsley: Oh, I saw how you update it in omnical_metrics_run. Could we maybe make an optional input the the init function? It's a bit odd to have the user populate an attribute like this manually after the fact.

hera_qm/omnical_metrics.py
```diff
@@ -12,3 +28,796 @@ def get_omnical_metrics_dict():
...
+    load firstcal delays and turn into phase gains
+
+    fc_file : str
+        path to firstcal .calfits file
+    """
+    uvf = UVCal()
+    uvf.read_calfits(fc_file)
+    freqs = uvf.freq_array.squeeze()
+    firstcal_delays = np.moveaxis(uvf.delay_array, 2, 3)[:, 0, :, :, :]
+    firstcal_gains = np.array(map(lambda x: np.exp(-2j*np.pi*freqs.reshape(1,-1,1)*x), firstcal_delays))
```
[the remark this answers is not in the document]
nkern: the `UVCal.convert_to_gain` function doesn't work for me. I think its the same issue zaki is having. 
the error is:
```
ValueError: UVParameter _total_quality_array is not expected shape. Parameter shape is (1, 1, 56, 1), expected shape is (1, 1024, 56, 1).
```

hera_qm/omnical_metrics.py
nkern: the plotting routines that live outside the class can be used with a generic `metrics` dictionary output. the plotting routines that live inside the class require data that doesn't come with a generic `metrics` dictionary, and therefore require one to make an instantiation of `OmniCal_Metics`. 

For example, the `plot_gains` function requires the `self.gain_diff` array, which is a very large array, so I don't put it into the `metrics` dictionary.

nkern:
@adampbeardsley let me know if you see anything else that needs to be addressed before we merge. perhaps tomorrow's telecon is a good time to see if there is anything else that @plaplant needs for the RTP.

hera_qm/omnical_metrics.py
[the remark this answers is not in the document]
adampbeardsley: Weird. I'm not sure if this is related to Zaki's issue or not. Could you document this in the pyuvdata issue log? Either comment on the open issue, or start a new one.

hera_qm/omnical_metrics.py
```diff
@@ -1,14 +1,872 @@
...
+
+        # Instantiate Data Object
+        self.uv = UVCal()
+        self.uv.read_calfits(omni_calfits)
+
+        # Get relevant metadata
+        self.Nants = self.uv.Nants_data
+        self.freqs = self.uv.freq_array.squeeze()
+        self.Nfreqs = len(self.freqs)
+        self.jones = self.uv.jones_array[0]
```
adampbeardsley: Is there an implicit assumption that only one polarization is present? We made this assumption for firstcal_metrics because we could get away with it in the RTP. But omnical files have multiple pols, and I don't think this will work.
Also, the M&C needs the pols to be antenna feed pols (single character `'x'` or `'y'`). We can work around this if there is need to include cross pol metrics, but my guess is that it's more accurate to keep it single.

hera_qm/omnical_metrics.py
```diff
@@ -1,14 +1,872 @@
...
+        metrics['ant_phs_noise'] = None
+        metrics['tot_phs_noise'] = None
+        metrics['phs_noise_bad_ants'] = None
+        metrics['phs_noise_good_sol'] = None
+        metrics['phs_noise_cut'] = None
+        metrics['ant_phs_std'] = None
+        metrics['tot_phs_std'] = None
+        metrics['phs_std_bad_ants'] = None
+        metrics['phs_std_good_sol'] = None
+        metrics['phs_std_cut'] = None
```
adampbeardsley: I just realized these `None`s may also lead to trouble with the M&C because the column requires floats. We could either come up with an obvious "this wasn't actually run" value (-1?), or in metrics2mc we can trim them out if their value is None.

hera_qm/omnical_metrics.py
nkern: yes, I guess that was how I interpreted your previous comment on keeping things separated, which I took to mean run the metrics on a file-per-file basis, in which case only one pol is present. let me know if you'd like me to switch back to the multi-pol version I had before, if you mean something entirely different.

hera_qm/omnical_metrics.py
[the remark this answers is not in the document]
nkern: based off of what zaki is now saying it seems likely it is the same issue. perhaps we should wait to see if the fix for zaki's problem also works for me.

hera_qm/omnical_metrics.py
adampbeardsley: Sorry for the confusion. The omnical files that the RTP produces have multiple pols in them. So there really isn't an option to run on a single pol file. I meant that we need to either make the module handle multiple pols - retaining which pol the metric is coming from so we can pass that on to the m&c, or if the metrics combine pols, we need to repeat the entry in the m&c so both antenna feeds show up with the same metric (ant_metrics does this for the cross pol metrics). 
One other solution would be to have a pol argument that the user supplies, then the module only uses that single pol. Then the script would be run twice - once with `'x'` and once with `'y'`.

nkern: ah, so what you are saying is that the RTP does not produce a `zen.JD.JD.xx.HH.uvc.omni.calfits` file but a `zen.JD.JD.HH.uvc.omni.calfits` file with all 4-pol in one file? If that's the case then I can go back to the multi-pol version I had, and I'll try and make sure the relevant metadata gets propagated to M&C

hera_qm/omnical_metrics.py
nkern: ok, I put in some if statements to trim them out in `metrics2mc` if they are None

hera_qm/omnical_metrics.py
adampbeardsley: Right - just one file with all pols. Well, I think right now it's just a two-pol file (xx and yy), but that may change depending on the result of an ongoing discussion of the output of omnical ( @plaplant can correct me if I'm wrong).

nkern: @adampbeardsley so i'm not entirely sure how to cleanly take a cross pol (i.e. 'xy' or 'yx') chisq and assign it to a single pol 'x' or 'y' per antenna, so i'm thinking of only using autopol (i.e. 'xx' or 'yy') to make the per-antenna metrics (like 'ant_chisq_avg' or 'ant_chisq_std' etc). do you have any thoughts on this?

adampbeardsley: I think that's a fair thing to do.

nkern:
@adampbeardsley you mentioned that the RTP is condensing all cross-pol `omni.calfits` outputs into a single `omni.calfits` file. Is it doing the same thing with `first.calfits` files? If so that will probably affect how I setup the `omnical_metrics_run.py` script.

adampbeardsley:
Looking at recent librarian files, it looks like firstcal produces independent files for the polarizations, but omnical produces one file for the two pols. I believe this is because firstcal only _uses_ one pol at a time, so it's more efficient for the RTP to launch two separate jobs, while omnical requires all pols to be present, so it's one big job.

nkern:
@adampbeardsley @plaplant So with the exception of the `pyuvdata` error that is breaking Travis, I think this is essentially ready to be merged. Once we fix the problem w/ `pyuvdata` we can let Travis run, but all tests check out for me locally.

hera_qm/utils.py
```diff
@@ -251,24 +272,48 @@ def metrics2mc(filename, ftype):
...
+            # pack antenna metrics, only uses auto pol (i.e. XX or YY)
+            if pol not in ['XX', 'YY']:
+                continue
+            cat = 'omnical_metrics_'
+            for met in ['chisq_ant_avg', 'chisq_ant_std', 'ant_phs_std']:
+                catmet = cat + met
+                try:
+                    if metrics[met] is None:
+                        continue
+                    d['ant_metrics'][catmet] = [[a, metrics['ant_pol'].lower(), metrics[met][a]] for a in metrics[met]]
```
adampbeardsley: This overwrites the previous polarization. For example, when I ran this with an omnical file with xx and yy, the yy iteration of this loop overwrite the xx iteration, and the output dictionary only had yy. There needs to be a check if `d['ant_metrics'][catmet]` already exists to append, rather than overwrite.

nkern: ah yes, that was dumb.. thanks

hera_qm/omnical_metrics.py
```diff
@@ -166,10 +166,12 @@ def load_firstcal_gains(fc_file):
     uvf = UVCal()
     uvf.read_calfits(fc_file)
     freqs = uvf.freq_array.squeeze()
-    firstcal_delays = np.moveaxis(uvf.delay_array, 2, 3)[:, 0, :, :, :]
-    firstcal_gains = np.array(map(lambda x: np.exp(-2j * np.pi * freqs.reshape(1, -1, 1) * x), firstcal_delays))
+    fc_gains = np.moveaxis(uvf.gain_array, 2, 3)[:, 0, :, :, :]
+    d_nu = np.mean(freqs[1:]-freqs[:-1])
+    d_phi = np.abs(np.mean(np.angle(fc_gains)[:, :, 1:, :] - np.angle(fc_gains)[:, :, :-1, :], axis=2))
+    fc_delays = (d_phi / d_nu)/(2*np.pi)
```
adampbeardsley: This code is now repeated in firstcal_metrics and omnical_metrics. We should move it into a utility function, but it's fine if we wait until after this PR.

PR #81, opened 2017-09-27
add functionality to load in delays from json file instead of first.c…

zakiali:
…alfits.

Added tests for testing that functionality and add new data files


hera_qm/firstcal_metrics.py
```diff
@@ -312,12 +312,33 @@ def __init__(self, calfits_file):
...
+        except KeyError:
+            raise ValueError('Sorry, only calibration polarizations "x" and '
+                             '"y" are currently supported.')
+
+        # Get the firstcal.rotated_antenna.metric file
+        if self.UVC.cal_type == 'gain':
+            with open(calfits_file + '.rotated_metric.json') as rot_file:
+                rmetric = json.load(rot_file)
+            self.delays = np.array([rmetric['delays'][str(ai)] for ai in self.UVC.ant_array]).squeeze()
+            self.rotated_antennas = { int(k): self.pol for k in rmetric['rotated_antennas'] }
```
adampbeardsley: Is there a way to add a check that the pol from the calfits file is the same as the pol for the rotated ants? Even better would be a check that they're the same observation.

jsdillon:
@nkern The current decision is not to track the rotated_antennas json, so please make the firstcal metrics json copy that info over. Not sure whether you do this yet...a bit hard to tell from skimming your code.

hera_qm/firstcal_metrics.py
nkern: I think it would be more robust to *not* make `firstcal_metrics` dependent on a `.json` file in order to run if the `cal_type` is == `gain`, which seems to be the standard from here on out since `hera_cal.firstcal` is outputting solutions with gains not delays. One thing that is easy is to just make `firstcal_metrics` calculate the delays if given the gains, which is pretty straightforward. Here is some code to do that, which matches exactly (to float precision) of the `delays` in the `.json` file. @zakiali I can insert this into this line here, or if you don't want merge conflicts you could do it yourself. Let me know what you think.
```
    freqs = self.UVC.freq_array.squeeze()
    fc_gains = np.moveaxis(self.UVC.gain_array, 2, 3)[:, 0, :, :, :]
    d_nu = np.mean(freqs[1:]-freqs[:-1])
    d_phi = np.abs(np.mean(np.angle(fc_gains)[:, :, 1:, :] - np.angle(fc_gains)[:, :, :-1, :], axis=2))
    fc_delays = (d_phi / d_nu)/(2*np.pi)
```

jsdillon: I like this idea, though I have a few concerns. 

First off, use np.diff rather than `freqs[1:]-freqs[:-1]`

Second, instead of mean, use median, which protects you from phase wraps. Alternatively, use unwrap. Or maybe do both?

Third, I think the abs gives you all positive delays...but some delays are negative!

Fourth, you also need to automatically detect the offsets in addition to the delays.

Finally, I think it's important that the delays and offsets are stored in an easily readable way somewhere...at the very least in the firstcal metrics json.

nkern: @jsdillon yes you are right, thanks, here is updated block:
```
fc_gains = np.moveaxis(self.UVC.gain_array, 2, 3)[:, 0, :, :, 0]
fc_phi = np.unwrap(np.angle(fc_gains))
d_nu = np.mean(np.diff(freqs))
d_phi = np.mean(fc_phi[:, :, 1:] - fc_phi[:, :, :-1], axis=2)
fc_delays = (d_phi / d_nu)/(-2*np.pi)
```

nkern: @jsdillon about the offsets: the json file that zaki writes out doesn't contain the offsets: is that a new capability you'd like `firstcal_metrics` to store, or are you referring to something else?

nkern:
@zakiali @jsdillon I have a push to this branch that fixes the interface problems w/ M&C and slightly reorganizes how the rotated antenna metrics are propagated through to firstcal_metrics.json. Let me know if you'd like me to push these changes to this branch so you can see them.

hera_qm/firstcal_metrics.py
jsdillon: Offsets is the same as the notion of rotated antennas. Rotated antennas have pi or -pi offsets, non-rotated antennas have 0 offset (e.g. the delay at freq=0). The firstcal_metrics needs to record rotated antennas, which are derivable from the gains, since the gains will +/- 1 * e^(-2 * pi * i * tau * nu) depending on whether or not the antenna is detected by firstcal as rotated.

jsdillon: Also, I feel like the median rather than the mean is more robust if something goes wrong with the unwrapping.

zakiali: Yea, this sounds good to me. Feel free to merge it in to this PR.

nkern:
@adampbeardsley so @zaki and I have reviewed this merge to `hera_qm` and once it passes travis we think it is good to go.

hera_qm/firstcal_metrics.py
```diff
@@ -312,15 +314,33 @@ def __init__(self, calfits_file):
...
+            raise ValueError('Sorry, only single pol firstcal solutions are '
+                             'currently supported.')
+        pol_dict = {-5: 'x', -6: 'y'}
+        try:
+            self.pol = pol_dict[self.UVC.jones_array[0]]
+        except KeyError:
+            raise ValueError('Sorry, only calibration polarizations "x" and '
+                             '"y" are currently supported.')
+
+        # Get the firstcal.rotated_antenna.metric file
```
adampbeardsley: Misplaced comment?

hera_qm/firstcal_metrics.py
```diff
@@ -406,8 +426,23 @@ def run_metrics(self, std_cut=0.5):
...
+            try:
+                with open(rotant_json) as rot_file:
+                    rmetric = json.load(rot_file)
+                self.rotated_antennas = np.array(rmetric['rotated_antennas']).astype(np.int).tolist()
+                metrics['rot_ants'] = self.rotated_antennas
+            except IOError:
+                print("couldn't load {0}".format(rotant_json))
+                metrics['rot_ants'] = None
+        else:
+            metrics['rot_ants'] = None
```
adampbeardsley: I'm confused. I thought the conclusion from the conversation on this PR was that firstcal_metrics could calculate the `rot_ants` on its own, and there really isn't a need for the json to go between. If all the information is there, I think it's much preferable to not use this intermediate json.

jsdillon: Agreed. It's not good enough it just get the delays with fc metrics. We also need to measure the rotations.

nkern:
@adampbeardsley @jsdillon sorry didn't totally understand. anyways, update has been pushed so that `firstcal_metrics` only needs a `.first.calfits` file and it calculates delay offsets and finds rotated antennas. delay offsets are stored in the firstcal metrics json but aren't propagated to M&C, only `rot_ants` flag is propagated to M&C. let me know if you see anything else! thanks

adampbeardsley:
This looks good. I just want to give @plaplant a chance to do integration tests before we merge.

nkern:
@adampbeardsley @plaplant actually wait just a sec.. i realized that `rot_ants` can be set to `None` which @adampbeardsley raised before could mess up `metrics2mc`. let me fix that and add a test and i'll push

plaplant:
It looks like this branch is running smoothly on RTP. Feel free to merge it in.

PR #82, opened 2017-09-28
Xrfi run revamp

adampbeardsley:
This PR rearranges how we run xrfi (`xrfi_run`) to allow for additional calibration and model vis files to flag on. It also introduces `xrfi_apply` to apply the calibration after the fact. The RTP will use these functions in the following way:

Each pol thread will run `xrfi_run`, but only one thread will pass in the cal and model files (because they contain all pols). Those processes will store .npz files with the flag_array and a thresholded waterfall.
Once all the pols are done, they can each run `xrfi_apply` to apply the union of all waterfalls to their respective data files. `xrfi_apply` has an option for `flag_file` as well as a list of waterfall files. `flag_file` will use the full flag array (e.g. the flag_array derived from the actual data), and the waterfalls will be unioned together and cast across the larger array.

There is one test that throws an annoying warning that I'd like to get rid of, but wanted to issue this PR to start the discussion.


hera_qm/tests/test_xrfi.py
```diff
@@ -308,145 +308,279 @@ def test_detrend_medfilt(self):
...
+        arg2 = "--model_file=" + model_file
+        calfits_file = os.path.join(DATA_PATH, 'zen.2457698.40355.xx.HH.uvcAA.omni.calfits')
+        arg3 = "--calfits_file=" + calfits_file
+        xx_file = os.path.join(DATA_PATH, 'zen.2457698.40355.xx.HH.uvcAA')
+        arguments = ' '.join([arg0, arg1, arg2, arg3])
+
+        arg4 = "--model_file_format=miriad"
+        cmd = ' '.join([arguments, arg4, xx_file])
+        args = a.parse_args(cmd.split())
+        nt.assert_raises(RuntimeError, xrfi.xrfi_run, args.filename, args, cmd)
```
plaplant: I think this is the line that's causing the error in the tests. (Locally, my nosetests crashes hard and won't finish running the rest of the tests.) I think the 3 solutions to this are: 1) We don't support Miriad as a format for the model visibilities, 2) We support Miriad as a format, but remove this test (so we don't cover the `uvm.read_miriad` line), or 3) convert the model visibility file from uvfits to Miriad and use that for the test.

I think for me, 3) is the most appealing solution, and can probably be done on-the-fly as part of the test. Then the Miriad model visibility file can be removed at the end of the test along with the test outputs. I'm open to other solutions too.

hera_qm/tests/test_xrfi.py
```diff
@@ -308,145 +308,279 @@ def test_detrend_medfilt(self):
...
         if os.path.exists(dest_file):
             shutil.rmtree(dest_file)
         cmd = ' '.join([arguments, xx_file])
         args = a.parse_args(cmd.split())
-        xrfi.xrfi_run(args.files, args, cmd)
+        history = cmd
+        xrfi.xrfi_apply(args.filename, args, cmd)
         nt.assert_true(os.path.exists(dest_file))
-        # clean up
-        shutil.rmtree(dest_file)
```
plaplant: We should probably still be cleaning up our output files.

hera_qm/tests/test_xrfi.py
```diff
@@ -308,145 +308,279 @@ def test_detrend_medfilt(self):
...
+        arg2 = "--extension=R.uvfits"
+        dest_file = os.path.join(DATA_PATH, 'test_output', 'zen.2457698.40355.xx.HH.uvcAAR.uvfits')
+        if os.path.exists(dest_file):
+            os.remove(dest_file)
+        arguments = ' '.join([arg0, arg1, arg2, arg3, arg4, arg5])
+        cmd = ' '.join([arguments, xx_file])
+        args = a.parse_args(cmd.split())
+        history = cmd
+        xrfi.xrfi_apply(args.filename, args, cmd)
+        nt.assert_true(os.path.exists(dest_file))
```
plaplant: Output file should be removed here too.

hera_qm/xrfi.py
```diff
@@ -225,145 +226,284 @@ def xrfi(d, f=None, Kt=8, Kf=8, sig_init=6, sig_adj=2):
...
-def xrfi_run(files, args, history):
+def xrfi_run(filename, args, history):
     """
-    Run an RFI-flagging algorithm on an entire file and store results in flag array.
+    Run an RFI-flagging algorithm on a single data file, and optionally calibration files,
+    and store results in npz files.
 
     Args:
-       files -- a list of files to run RFI flagging on
+       filename -- Data file to run RFI flagging on. Single polarization is assumed.
```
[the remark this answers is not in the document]
adampbeardsley: Good point. It's actually not necessary, just what I was thinking about with respect to RTP flow. I'll remove it.

adampbeardsley:
@plaplant I think I've addressed your comments if you want to take another look.

plaplant:
Changes look great! The travis issues look like they're related to the antenna diameters fix being addressed in another branch, so as soon as the tests pass this is good to go.

Issue #84, opened 2017-10-03
Remove "too many figures" warning from tests


plaplant:
This is very low priority, but on travis, we are seeing a warning along the lines of `More than 20 figures have been opened` related to our matplotlib tests. This can be solved by explicitly closing our plots when we're finished with them, using syntax like `plt.close(fig)`. If the figure objects are being passed back to the user, then they should be closed in the tests. If not, they should be closed in the modules.

plaplant:
This was addressed in PR #88.

PR #88, opened 2017-10-03
Close figures after saving them

plaplant:
This PR addresses Issue #84. The warning appears to have gone away from Travis, which means things are more memory-friendly for end-users as well.

### DIFF
--- a/hera_qm/firstcal_metrics.py
+++ b/hera_qm/firstcal_metrics.py
@@ -163,6 +163,9 @@ def plot_stds(metrics, fname=None, ax=None, xaxis='ant', kwargs={}, save=False):
             fname = metrics['fc_filestem'] + '.stds.png'
         fig.savefig(fname, bbox_inches='tight')
 
+        # close figure
+        plt.close(fig)
+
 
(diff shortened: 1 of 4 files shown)


nkern:
@plaplant To fix the tests, I've just been doing `plt.close()` after a test involving a plotting routine to close all open plots, which has solved the travis issue in the past.

One might think of a situation where I'm using a notebook to make a plot, but also want to write the plot out to file w/ the `save` keyword at the same time. If we close the figure in the script, we won't be able to do both at the same time. 

One can think of an easy way around this (just run the script twice, one w/ `save=False` and `save=True'). I'm fine w/ this fix but just something to consider...

plaplant:
If it's easier to just add a `plt.close()` statement to the tests, that's fine with me. I'll roll back these changes and try that instead.

nkern:
@plaplant yeah I did it in `test_omnical_metrics`, but I must have forgotten to do it in `test_firstcal_metrics`. thanks

plaplant:
@nkern this PR should be ready for another look

nkern:
@plaplant looks good!

PR #90, opened 2017-10-04
Fc metrics add multiple files

nkern:
enabled `firstcal_metrics.FirstCal_Metrics` to take multiple files

along the same lines, enabled subtraction of long-time delay modes from delay solutions via gaussian process regression before taking standard deviation. this means `hera_qm` now has a
`scikit-learn` dependency, but it is a weak dependency in that `hera_qm` will still operate normally if `sklearn` can't be imported

also updated some tests related to a previous rotated antenna PR that got left behind...


hera_qm/firstcal_metrics.py
```diff
@@ -279,12 +285,21 @@ class FirstCal_Metrics(object):
...
+    sklearn_import = sklearn_import
+
+    def __init__(self, calfits_files, use_gp=True):
         """
         Input:
         ------
-        calfits_file : str
+        calfits_files : str or list
             filename for a *.first.calfits file
+            or a list of .first.calfits files
```
adampbeardsley: Please clarify here that the list of files must still only be one polarization.

hera_qm/tests/test_firstcal_metrics.py
```diff
@@ -188,20 +196,6 @@ def test_firstcal_metrics_run(self):
         self.assertTrue(os.path.exists(dest_file))
         os.remove(dest_file)
 
-        # test w/ rotant json
```
adampbeardsley: In the future let's try to keep branches/PRs limited to a single idea, and make a separate PR for handling other stuff that's found in the process.

hera_qm/firstcal_metrics.py
```diff
@@ -366,6 +385,25 @@ def __init__(self, calfits_file):
         self.delay_avgs = np.median(self.delays, axis=1)
         self.delay_fluctuations = (self.delays.T - self.delay_avgs).T
 
+        # use gaussian process model to subtract underlying mean function
+        if use_gp is True and self.sklearn_import is True:
+            kernel = gp.kernels.RBF(length_scale=0.2, length_scale_bounds=(0.01, 1.0)) + gp.kernels.WhiteKernel(noise_level=0.01)
```
adampbeardsley: Some of the hard-coded values here are a little opaque. Could you add comments explaining them and/or derive them programmatically?

PR #92, opened 2017-10-04
fc_metrics : return figure when custom_ax is False

nkern:
a quick fix to allow the plotting routines to return the figure object when a figure and axis is created on-the-fly. helpful when plotting in a notebook...


nkern:
also did same for plotting routines in `omnical_metrics`, despite branch name.. let me know if you'd rather me do this in a separate PR...

PR #94, opened 2017-10-05
added firstcal metric max_std

nkern:
also changed firstcal good / bad sol metric to be dependent on `max_std` rather than `agg_std`

this is a stronger metric b/c even if one antenna is bad, once omnical runs, the bad antenna will affect the solutions of all the other antennas


adampbeardsley:
Are antennas that firstcal_metrics IDs as "bad" flagged for omnical? If so, wouldn't it be better to use the max of antennas that are not flagged for good/bad sol?

nkern:
as far as i know, firstcal bad_ants flags do not get propagated to omnical

adampbeardsley:
huh. Ok, maybe we should talk about that on the hera_cal side.

PR #97, opened 2017-10-05
Add in rms plot in auto_view.py

zakiali:
Iv'e added in the a function for the plot. Let me know if that function should live somewhere else other than the script itself. Also, if there are any style modifications to be made.


zakiali:
This is the generated plot for one polarization 
zen 2458001 11719 xx auto_rms_values

nkern:
@zakiali might be useful to put in the title which polarization in that case.. and perhaps make antenna text a color other than black..

zakiali:
better?
zen 2458001 11719 xx auto_rms_values

adampbeardsley:
@nkern Did you have any other comments on this one before we merge?

PR #98, opened 2017-10-05
ensures omnical_metrics.load_firstcal_gains matches polarization of .omni.calfits

nkern:
enables omnical_metrics.OminCal_Metrics.load_firstcal_gains to reorder polarization of incoming fc_files if it doesn't match ordering of omni.calfits file.


adampbeardsley:
Why are the test calfits files changing?

nkern:
because in previous PR's i accidently overwrote them and committed them. the only one's Ive changed are the uvcA.*42234.*calfits files. the only difference was the antenna number, which should be 17. previously I excluded antenna 43, which we need to have in the file so that our rotated_antenna metric can pick it up

PR #104, opened 2017-10-13
Add ex_ants options

plaplant:
This PR adds an `ex_ants` and `metrics_json` command line options, similar to those in `omni_run` from hera_cal. `ex_ants` allows the user to specify a comma-separated list of antennas to exclude from RFI threshold calculations, and `metrics_json` allows the user to pass in a JSON file, where it is assumed one of the dictionary keys is `ex_ants`. The visibilities corresponding to these antennas are flagged in the specified visibility before the xrfi flagging algorithm is run, and so the flags are normalized out in thresholding calculations.

One drawback of this approach is that this functionality is added to `xrfi_run`, and as such, the flags are _not_ saved back out to the input visibility file. If that functionality is desired, I think we should add that in the `xrfi_apply` step, which is a more natural spot for it. I could add that in without too much trouble, but thought I'd make the PR now to see what other people think.

### DIFF
--- a/hera_qm/tests/test_xrfi.py
+++ b/hera_qm/tests/test_xrfi.py
@@ -519,6 +519,30 @@ class TestXrfiRun(object):
         # clean up
         os.remove(dest_file)
 
+    def test_xrfi_run_exants(self):
+        # get argument object
+        a = utils.get_metrics_ArgumentParser('xrfi_run')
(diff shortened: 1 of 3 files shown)


adampbeardsley:
I think the functionality as is is perfect. We don't want to change the input file, and as you have it, the ex_ants flags will propagate to the output file. So no need to add anything.

PR #109, opened 2017-10-19
Fix flag summary

adampbeardsley:
This PR fixes a bug where an "alternative flag array" was not actually being used when summarizing flags. It also introduces prior flags which can be excluded when calculating averages.


hera_qm/tests/test_xrfi.py
```diff
@@ -688,6 +688,28 @@ def test_summarize_flags(self):
         self.assertEqual(data['version'], hera_qm_version_str)
         os.remove(outfile)  # cleanup
 
+    def test_summarize_flags_with_prior(self):
+        from hera_qm.version import hera_qm_version_str
```
plaplant: Looks like this line is not needed.

PR #114, opened 2018-02-14
Fix to watershed algorithm when looking for outliers next to flagged channels.

zakiali:
In my attempts to RFI flag I found a bug in the watershed algorithm as written. Once the main flags have been found, it looked like the flagging of possible RFI next to neighboring channels was not working well. It was mistakenly using flags as the sigma array instead of the actual sigma array when comparing to sig_adj. If sig_adj = 1, nothing changes, but any other value would over or underestimate the flag array.

### DIFF
--- a/hera_qm/tests/test_xrfi.py
+++ b/hera_qm/tests/test_xrfi.py
@@ -901,17 +901,18 @@ class TestFlagXants(object):
                 flags = uv.flag_array[blts, :, :, :]
                 nt.assert_true(np.allclose(flags, True))
 
+
 class TestInputFlagWatershed(object):
     def test_input_flag(self):
         # create some fake data for input to watershed
-        SIZE=10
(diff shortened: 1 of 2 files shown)


adampbeardsley:
Oh interesting. It looks like if you don't pass a priori flags, it's identical (because `f1` becomes a masked array of `d`), but if you do pass something in, it won't do anything useful. Good catch.
Could you:
1) Add a quick test to demonstrate this is doing what is should do
2) Get travis to pass. It looks like hera_cal requires linsolve now, so our build isn't working. Normally this would be a new issue, but you're a pretty good candidate to address hera_cal dependencies anyway, so let's see if we can do it all in one go.

adampbeardsley:
Looks like Josh fixed the travis issue in his recent PR. So if you merge in the changes from master, it should pass.

Issue #113, opened 2018-02-14
Speed up ant_metrics


plaplant:
The largest bottleneck in RTP currently is the ant_metrics stage of processing. We should make this step take less time to better fit within the "real-time" constraint of the system.

@jsdillon I'm tentatively assigning this to you, but feel free to reassign as necessary. Thanks for looking into this.

jsdillon:
The simplest minor fix is to just throw away antennas that appear really dead (dead_metric > 10) whenever they appear, even if they aren't the worst offenders. This eliminates many of the iterations of ant_metrics, though not all.

I've verified on a day of ant_metrics data that this doesn't throw away any extra antennas that weren't going to get thrown out eventually.

```
import glob
amfiles = glob.glob('../../hera_sandbox/jsd/Antenna_Metrics/sample_jsons/*')
ams = [ant_metrics.load_antenna_metrics(f) for f in amfiles]

starts_below_10 = set([])
starts_above_10 = set([])
ends_above_5 = set([])
ends_below_5 = set([])

for n,am in enumerate(ams):
    for ant in am['final_mod_z_scores']['meanVij'].keys():
        final_dead_metric = np.abs(am['final_mod_z_scores']['meanVij'][ant]/2) + np.abs(am['final_mod_z_scores']['redCorr'][ant]/2)
        starting_dead_metric = np.abs(am['all_mod_z_scores'][0]['meanVij'][ant]/2) + np.abs(am['all_mod_z_scores'][0]['redCorr'][ant]/2)
        if final_dead_metric > 5:
            ends_above_5.add(ant+(n,))
        if final_dead_metric <= 5:
            ends_below_5.add(ant+(n,))
        if starting_dead_metric > 10:
            starts_above_10.add(ant+(n,))
        if starting_dead_metric <= 10:
            starts_below_10.add(ant+(n,))

print 'starts_below_10', len(starts_below_10)
print 'starts_above_10', len(starts_above_10)
print 'ends_below_5', len(ends_below_5)
print 'ends_above_5', len(ends_above_5)

print 'starts above 10 and ends below 5:', len(starts_above_10.intersection(ends_below_5))
print 'starts below 10 and ends above 5:', len(starts_above_10.intersection(ends_above_5))
```
which yields
```
starts_below_10 9000
starts_above_10 216
ends_below_5 8889
ends_above_5 327
starts above 10 and ends below 5: 0
starts below 10 and ends above 5: 216
```

[...]

PR #115, opened 2018-02-15
Speed up ant_metrics a bit with a notion of an "always dead" cut

jsdillon:
ant_metrics, as it's currently designed, only throws away one antenna at a time. This is slow if we have to throw away a lot of antennas. 

This somewhat hacky solution allows ant metrics to kill multiple dead antennas if they are all "very dead." I've defaulted the very dead to 10 sigma (double the 5 sigma cut for dead). On each iteration, if the worst antenna is dead (not crossed), then it will throw away all antennas that are very dead (including, of course, the worst dead antenna).

This addresses #113  In my testing, the "always cut" of 10 finds about 2/3 of the antennas that will eventually be flagged but never flags an antenna that wouldn't eventually be flagged anyway. This should get us a factor of 2-3 speed-up, which should be fine for H1C. Let's revisit all of this for H2C.


jsdillon:
ok, @adampbeardsley @plaplant got the tests to pass. Ready for you guys to take a look.

hera_qm/utils.py
```diff
@@ -31,6 +31,9 @@ def get_metrics_ArgumentParser(method_name):
                        help='Modified z-score cut for most cross-polarized antenna. Default 5 "sigmas"')
         a.add_argument('--deadCut', default=5.0, type=float,
                        help='Modified z-score cut for most likely dead antenna. Default 5 "sigmas"')
+        a.add_argument('--alwaysDeadCut', default=1.0, type=float,
```
plaplant: Should we have `default=10.` instead? That's what the help suggests.

jsdillon: Oops. My bad. Yes.

PR #117, opened 2018-02-20
Support reading metrics files with inf values

plaplant:
This PR supports reading `inf` values saved in metrics files. Right now they break the file reader, and we should support loading them (even if they signify bad data or incorrect processing). This also introduces a test specifically for the `load_antenna_metrics` function.


plaplant:
Yes, I ran some local tests on the `load_firstcal_metrics` and `load_omnical_metrics` functions, and they can handle `np.nan` and `np.inf` without a problem. If we ever make a unified "metrics" base class, we should use the read/write functions from the latter as a model. I can push the firstcal/omnical tests up if you want, but they're not terribly illuminating (except to convince ourselves it's Doing The Right Thing).

PR #119, opened 2018-02-20
XRFI Simple Bug Fix

jkerrigan:
Fixed a bug (or feature?) where xrfi_simple would automatically flag the waterfall visibility edges, however functionality without this appears the same minus the automatically flagged beginning and end time samples across all frequencies.


adampbeardsley:
Hmm, this is curious. It would be good to have an idea why it was there to begin with. Best I can tell @AaronParsons put it in there in the Before Time. Maybe he can tell us whether it should remain?

adampbeardsley:
Given that we're burning down xrfi_simple in the xrfi_revamp, I'm going to close this.

PR #124, opened 2018-03-09
Xrfi after delay filter

adampbeardsley:
This PR enables the delay_xrfi_run script which does a delay filter before rfi flagging. It does a few things:
- allows a UVData object to be passed to xrfi_run
- adds an output npz in xrfi_apply to save the flags actually applied to the data (separately)
- creates the delay_xrfi_run script

### DIFF
--- a/hera_qm/utils.py
+++ b/hera_qm/utils.py
@@ -150,7 +150,9 @@ def get_metrics_ArgumentParser(method_name):
                        'list of npz files containing waterfalls of flags to broadcast '
                        'to full flag array and union with flag array in flag_file.')
         a.add_argument('--output_npz', default=True, type=bool,
-                       help='Whether to save an npz with the final flag array and waterfall.')
+                       help='Whether to save an npz with the final flag array and waterfall. '
+                       'The flag array will be identical to what is stored in the data, '
+                       'and the waterfall is the union of all input waterfalls.')
         a.add_argument('--out_npz_ext', default='.flags.npz', type=str,
                        help='Extension to be appended to input file name. Default is ".flags.npz".')
         a.add_argument('filename', metavar='filename', nargs='*', type=str, default=[],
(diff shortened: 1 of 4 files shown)


hera_qm/utils.py
```diff
@@ -149,6 +149,10 @@ def get_metrics_ArgumentParser(method_name):
         a.add_argument('--waterfalls', default=None, type=str, help='comma separated '
                        'list of npz files containing waterfalls of flags to broadcast '
                        'to full flag array and union with flag array in flag_file.')
+        a.add_argument('--output_npz', default=True, type=bool,
```
plaplant: Is this the "decision waterfall" that we were talking about on the telecon? If it is, then maybe make the docs explicit to reference this.

hera_qm/xrfi.py
```diff
@@ -249,37 +249,44 @@ def xrfi(d, f=None, Kt=8, Kf=8, sig_init=6, sig_adj=2):
...
-    if args.infile_format == 'miriad':
-        uvd.read_miriad(filename)
-    elif args.infile_format == 'uvfits':
-        uvd.read_uvfits(filename)
-    elif args.infile_format == 'fhd':
-        uvd.read_fhd(filename)
+    if isinstance(indata, UVData):
+        uvd = indata
+        if len(args.filename) == 0:
+            raise AssertionError('Please provide a filename to go with UVData object.')
```
plaplant: This is a little counter-intuitive to me (why would we need the filename if we already have the data in memory?), so maybe add a comment that the filename is used to generate the name of output.

hera_qm/utils.py
adampbeardsley: Whoops, I was actually not even storing the waterfall itself. I'll add a better description, and store the waterfall (if applicable).

PR #125, opened 2018-03-13
Fix nans in antmetrics

adampbeardsley:
redcorr metrics should give NaNs for antennas with no redundant baselines.

I think how to treat this is still controversial, but I think this answer is better than the current implementation (zeros).


hera_qm/ant_metrics.py
```diff
@@ -135,6 +135,7 @@ def red_corr_metrics(data, pols, antpols, ants, reds, xants=[], rawMetric=False,
 
     # Compute power correlations and assign them to each antenna
     autoPower = compute_median_auto_power_dict(data, pols, reds)
+    red_ants = list(set([ant for r in reds for bl in r for ant in bl]))
```
plaplant: Is this list used for anything? I don't see it referenced later in the code.

adampbeardsley: Ah, no. I was going to use it, but then realized I could just use antCounts

PR #131, opened 2018-03-20
Delay xrfi a priori flags

adampbeardsley:
This PR goes hand in hand with one on hera_opm. The goal is to run xrfi on calfits and model vis files before running the delay filter + xrfi on data.
Biggest changes here are updating the top level script to do what is described above, allowing xrfi_run to not take an input file, and a new argument parser to handle this.

### DIFF
--- a/hera_qm/utils.py
+++ b/hera_qm/utils.py
@@ -5,17 +5,20 @@ import warnings
 import argparse
 import numpy as np
 
+
 # argument-generating function for *_run wrapper functions
 def get_metrics_ArgumentParser(method_name):
     """
     Function to get an ArgumentParser instance for working with metrics wrappers.
 
     Args:
-        method_name -- target wrapper, must be "ant_metrics", "firstcal_metrics", "omnical_metrics", or "xrfi"
+        method_name -- target wrapper, must be "ant_metrics", "firstcal_metrics",
+                       "omnical_metrics", "xrfi_run", "delay_xrfi_run", or "xrfi_apply"
     Returns:
         a -- an argparse.ArgumentParser instance with the relevant options for the selected method
     """
-    methods = ["ant_metrics", "firstcal_metrics", "omnical_metrics", "xrfi_run", "xrfi_apply"]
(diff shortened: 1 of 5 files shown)


hera_qm/utils.py
```diff
@@ -132,6 +135,75 @@ def get_metrics_ArgumentParser(method_name):
                        'visibilities formed with these antennas will be set to True.')
         a.add_argument('filename', metavar='filename', nargs='*', type=str, default=[],
                        help='file for which to flag RFI (only one file allowed).')
+    elif method_name == 'delay_xrfi_run':
```
jsdillon: Are you familiar with argument groups? e.g. https://github.com/HERA-Team/hera_cal/blob/7ee5af6186e278ef8cd627cc81b2853e3e62d88a/hera_cal/smooth_cal.py#L430

Might make this whole thing more intelligible.

adampbeardsley: I saw you doing that in hera cal. It's just an organizational thing right? Doesn't actually change behavior?
Are you suggesting using it within one of these "methods", or to replace the whole if/else block?

hera_qm/xrfi.py
```diff
@@ -265,7 +266,12 @@ def xrfi_run(indata, args, history):
     model visibility file, and run an RFI-flagging algorithm to identify contaminated
     observations. Each set of flagging will be stored, as well as compressed versions.
     """
-    if isinstance(indata, UVData):
+    if indata is None:
+        if (args.model_file is None) and (args.calfits_fils is None):
+            raise AssertionError('Must provide at least one of: filename, '
```
jsdillon: filename --> indata?

hera_qm/utils.py
jsdillon: It's purely organizational. And it makes the -h much neater.

adampbeardsley:
@jsdillon I added a test, which indeed caught a bug. Could you review again?

jsdillon:
oooh, look at that fun typo `calfits_fils`

Looks good to me.

PR #133, opened 2018-03-21
Fix calfits warnings

adampbeardsley:
Getting rid of warnings that get dumped during the tests.


adampbeardsley:
@nkern  would you mind reviewing this when you get a chance? All it does is replace calfits files with newer format so we don't get spammed with warnings.

Issue #134, opened 2018-03-21
Make xrfi_run more modular


adampbeardsley:
xrfi_run started as a simple concept, but gradually grew to handle all kinds of different run modes. We should revamp it to be simpler but allow more flexible input. For example, passing in model and calfits files in addition to data files is kind of silly. We should just allow for one input, and tell it what's being passed in.

jsdillon:
While you're doing this, can you add a test that exposes the error in #138 ? Thanks.

plaplant:
Fixed by #190.

PR #136, opened 2018-03-21
Another xrfi run fix

adampbeardsley:


jsdillon:
Also, this is definitely highlighting the need for #134

PR #137, opened 2018-03-22
changed defaults on delay filtering for xrfi

jsdillon:


jsdillon:
opm has already been updated

jsdillon:
@adampbeardsley looks like you have to review again :(

PR #138, opened 2018-03-22
abspath wasn't defined. Replaced it with os.path.abspath(args.model_file)

jsdillon:


plaplant:
Sorry, I know that time is of the essence, but could we quickly add a test that exposes this? It seems like a pretty big oversight that this bug made it through our testing.

jsdillon:
Agreed, but I'm going to make that an issue for Adam when he gets back (I think he said he's out of the country?).

Issue #140, opened 2018-04-03
Kernel size exceeds data in some cases


jsdillon:
See `/lustre/aoc/projects/hera/H1C_IDR2/IDR2_1/2458114/zen.2458114.66640.xx.HH.uv.CAL_XRFI.xy.log`

The simplest solution here is to just flag the whole file. If the issue is that it's very few integrations, another possible solution is to concatenate with the previous/next file and then use them together to find RFI.

adampbeardsley:
From the `hera_qm.xrfi` perspective, I think it makes most sense to flag the whole file. Concatenating with other files would have to be done on the analysis level, and maybe that's another avenue we can look at in parallel.

@jkerrigan Have we talked about this failure mode before? I vaguely remember this coming up, and we made a decision to have it error. But maybe I'm misremembering. What do you think about flagging all the data if your kernel size is bigger than the data?

jsdillon:
Flagging the whole file is fine with me.

PR #141, opened 2018-04-04
Have xrfi catch error when kernel is too large, flag all data.

adampbeardsley:
Addresses #140


hera_qm/tests/test_xrfi.py
```diff
@@ -71,7 +71,7 @@ def _run_test(self, func, arg, correct_flag, false_positive, nsig=5, fmode=False
                 f = func(data, *arg)
             except:
                 # ValueError check to make sure kernel size isn't too big
```
jsdillon: This comment is out of date.

adampbeardsley:
@jsdillon I updated the comment you pointed out, which dismissed your review. Could you re-review when you get a chance?

PR #145, opened 2018-04-30
added baseline_array to xrfi npz output

nkern:
This will help us to tie RFI flags across files without having to load in the visibility data itself. If we can index RFI flags based on their baseline, we can begin to broadcast flags across an entire night (similar to a `time_threshold` but for the whole night)

### DIFF
--- a/hera_qm/tests/test_xrfi.py
+++ b/hera_qm/tests/test_xrfi.py
@@ -344,6 +344,11 @@ class TestXrfiRun(object):
         history = cmd
         xrfi.xrfi_run(args.filename, args, cmd)
         nt.assert_true(os.path.exists(dest_file))
+        # load data and test all arrays exist in file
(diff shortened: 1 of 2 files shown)


adampbeardsley:
If we're going to be using this to tie flags across files, should we also have the time array? One can basically reconstruct it, but I worry about making implicit assumptions about the order of the data.

Alternatively, could you use the flag summary? That has all the info we currently use to do thresholding.

nkern:
Well, I was mainly using this to compute the total fraction of flags for a baseline-freq across all times, so the ordering of the times doesn't matter to me, and i didn't want to put more metadata than I needed. but if you think it would be useful to also have the time arrays i can put them in here. let me know what you think

adampbeardsley:
Oh, I see. You're doing a different cut than the summaries.

PR #146, opened 2018-05-02
Have threshold_flags flag places where prior was 100% flagged

adampbeardsley:
Now pior=1 results in normalization=1, instead of nan.


adampbeardsley:
I also just realized we may want to pass the prior flags into the thresholding to do a proper weighted average. But this will likely result in changing defaults... @jsdillon you can tell me if you think this should go into this PR. Otherwise I'll make an issue to address this at some point.

adampbeardsley:
Ugh, now I'm thinking this fix will already mess with those averages because we do nanmeans, which was down weighting the prior=1 pixels (because they got normalized to nan). Maybe the priors do need to get in there for this PR.

adampbeardsley:
Ok, I changed this a little. The normalization function is back to what we were doing previously (I kept it as a function now that it has a unit test). Now the threshold_flags function will explicitly flag pixels which are 100% flagged in the prior flags (ie, NaN in the normalized waterfall). This solution won't change any of the broadcasting, but it will propagate flags in a little better way.
*Needs to be revisited when we do the grand refactoring*
Ready for review @jsdillon

hera_qm/xrfi.py
```diff
@@ -555,6 +554,19 @@ def waterfall2flags(waterfall, uv):
     return flag_array
 
 
+def normalize_wf(wf, wfp):
+    """ Normalize waterfall to account for data already flaggedself.
```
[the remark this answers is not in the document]
adampbeardsley: Garr, stupid predictive typing.

adampbeardsley:
@jsdillon should we merge this (needs official approval)?
@plaplant did you want to take a quick look? I think you were involved in the original coding of this part.

plaplant:
Sure, I can take a quick look later today.

hera_qm/xrfi.py
```diff
@@ -557,6 +556,21 @@ def waterfall2flags(waterfall, uv):
...
+    Args:
+        wf -- Waterfall of fractional flags.
+        wfp -- Waterfall of prior fractional flags. Size must match wf.
+    Returns:
+        wf_norm -- Waterfall of fractional flags which were not flagged prior.
+                   Note if wfp[i, j] == 1, then wf_norm[i, j] will be NaN.
+    """
+    if wf.shape != wfp.shape:
+        raise AssertionError('waterfall and prior waterfall must be same shape.')
+    wf_norm = (wf - wfp) / (np.ones_like(wf) - wfp)
```
plaplant: See my main review comment. I think that we can be more clever here with a `where` statement, and have the output have a value of `1` instead  of `NaN` when `wfp == 1`.

hera_qm/xrfi.py
```diff
@@ -576,6 +590,7 @@ def threshold_flags(wf, px_threshold=0.2, freq_threshold=0.5, time_threshold=0.0
     tseries = np.nanmean(wf, axis=1)
     wf_t[tseries > freq_threshold, :] = True
     wf_t[wf > px_threshold] = True
+    wf_t[np.isnan(wf)] = True  # Flag anything that was completely flagged in prior.
```
plaplant: See comment above.

hera_qm/xrfi.py
adampbeardsley: I agree this is probably not the best solution. However, if possible I would like to carry the fact that `wf_norm` didn't actually know anything about these pixels. The idea is that if it was previously flagged, and we map that to a 1, it will mess up the thresholding done on lines 591 and 592. What would you think of this compromise:
`wf_norm = np.where(wfp < 1, (wf - wfp) / (np.ones_like(wf) - wfp), np.nan)`
This gets rid of the warnings, but retains the nans. Is that still poor form?

Issue #148, opened 2018-05-09
Add baseline, frequency, and time axes to output flag npzs


adampbeardsley:
To help the flag files be more independent, it has been requested that we store the baseline, frequency, and time arrays.

If this is urgent @jsdillon, I may have to ask @nkern to look at it (essentially an extension of his previous PR). I'm traveling, so won't be able to do anything immediately.

adampbeardsley:
Addressed in #149

PR #149, opened 2018-05-09
added metadata to flags.applied.npz

jsdillon:
Addresses #148


adampbeardsley:
Don't we want the full arrays (e.g. time, lst, baseline)? Otherwise you can't immediately use the full flag_array.

hera_qm/xrfi.py
```diff
@@ -735,7 +735,14 @@ def xrfi_apply(filename, args, history):
     else:
         raise ValueError('Unrecognized output file format ' + str(args.outfile_format))
     if args.output_npz:
-        # Save an npz with the final flag array and waterfall
+        # Save an npz with the final flag array and waterfall and relevant metadata
         outpath = outpath + args.out_npz_ext
-        np.savez(outpath, flag_array=uvd.flag_array, waterfall=wf_full,
-                 history=flag_history + history)
+        lsts = []
+        for l in uvd.lst_array.ravel():
```
zakiali: does np.unique not work for uvd.lst_array?

nkern: nope... that reorders it when it could be wrapped inherently. think about an lstarray that is [4, 4.5, 5, 5.5, 6, 0.25, ...]

jsdillon:
This is all just copy-pasted from hera_cal.io.load_vis so that I get all the same outputs (except the data of course).

jsdillon:
@adampbeardsley This is all just copy-pasted from hera_cal.io.load_vis so that I get all the same outputs (except the data of course)... this assumes that all baselines are there for all times... as does much of hera_cal.

adampbeardsley:
Also, should we be adding this to the other npzs that are saved in xrfi_run?

jsdillon:
Maybe... but this is what I need right now. 

Let's come back later to revisit the xrfi data products when we decide how to reorganize.

adampbeardsley:
@jsdillon even if you assume the baselines are all there, do you get the order you need?

jsdillon:
Good point... let me rethink this.

jsdillon:
I forgot that you can't still call .get_flags(). So now I'm doing as you suggest, @adampbeardsley. Thoughts?

jsdillon:
And thanks for a quick turnaround and finding my error.

PR #156, opened 2018-05-20
Normalize cal flags

adampbeardsley:
In `xrfi_run`, we neglected to use the normalize waterfall function on the gains, chi squareds, and model visibilities. Probably because in the past flags didn't exist from those stages. This PR adds the normalization in, and also fixes some handling of runtime warnings (which were causing the tests to fail when things were normalized properly).

### DIFF
--- a/hera_qm/xrfi.py
+++ b/hera_qm/xrfi.py
@@ -414,12 +414,12 @@ def xrfi_run(indata, args, history):
             dirname = os.path.dirname(os.path.abspath(args.calfits_file))
         outfile = ''.join([os.path.basename(args.calfits_file), '.g', args.extension])
         outpath = os.path.join(dirname, outfile)
-        np.savez(outpath, flag_array=g_flag_array, waterfall=g_wf_t, ants=uvc.antenna_numbers,
+        np.savez(outpath, flag_array=g_flag_array, waterfall=g_wf_t, ants=uvc.ant_array,
                  jones_array=uvc.jones_array, freq_array=uvc.freq_array,
                  time_array=uvc.time_array, history=history)
         outfile = ''.join([os.path.basename(args.calfits_file), '.x', args.extension])
         outpath = os.path.join(dirname, outfile)
-        np.savez(outpath, flag_array=x_flag_array, waterfall=x_wf_t, ants=uvc.antenna_numbers,
+        np.savez(outpath, flag_array=x_flag_array, waterfall=x_wf_t, ants=uvc.ant_array,
                  jones_array=uvc.jones_array, freq_array=uvc.freq_array,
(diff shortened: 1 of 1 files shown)


hera_qm/xrfi.py
```diff
@@ -399,18 +404,24 @@ def xrfi_run(indata, args, history):
...
+                 antpairs=uvm.get_antpairs(), polarization_array=uvm.polarization_array, freq_array=uvm.freq_array,
+                 time_array=uvm.time_array, lst_array=uvm.lst_array, antpos=antpos, ants=ants, history=history)
     if args.calfits_file is not None:
         # Save flags from gains and chisquareds in separate files
         if args.xrfi_path == '':
             dirname = os.path.dirname(os.path.abspath(args.calfits_file))
         outfile = ''.join([os.path.basename(args.calfits_file), '.g', args.extension])
         outpath = os.path.join(dirname, outfile)
-        np.savez(outpath, flag_array=g_flag_array, waterfall=g_wf_t, history=history)
+        np.savez(outpath, flag_array=g_flag_array, waterfall=g_wf_t, ants=uvc.antenna_numbers,
```
nkern: correct me if i'm wrong, but isn't it the `ant_array` that is technically what we want here?

hera_qm/xrfi.py
```diff
@@ -399,18 +404,24 @@ def xrfi_run(indata, args, history):
...
+                 jones_array=uvc.jones_array, freq_array=uvc.freq_array,
+                 time_array=uvc.time_array, history=history)
         outfile = ''.join([os.path.basename(args.calfits_file), '.x', args.extension])
         outpath = os.path.join(dirname, outfile)
-        np.savez(outpath, flag_array=x_flag_array, waterfall=x_wf_t, history=history)
+        np.savez(outpath, flag_array=x_flag_array, waterfall=x_wf_t, ants=uvc.antenna_numbers,
```
nkern: same here

hera_qm/xrfi.py
adampbeardsley: Yup, you're right. Good catch. I always get confused with these.

jsdillon:
@adampbeardsley @nkern is this good to go? I'd like to start a long run asap.

adampbeardsley:
I'm good with this. I was going to look at fine tuning thresholds, but I got swamped yesterday, and don't see any light in the next few days. I think the 1-day test is IDR2.1 worthy. @nkern ?

nkern:
Yeah I think so, perhaps @jsdillon we should discuss the long-run during the telecon in about 20 min and if we get a go-ahead consensus we start the pipeline soon after.

PR #171, opened 2018-07-05
Translated some exceptions to match pyuvdata update

adampbeardsley:

PR #172, opened 2018-07-12
Fixes to bring up to date with dependencies

adampbeardsley:
UVData._key2inds() now returns a tuple of pol inds to match ind1 and ind2. Just updated how we handle that in xrfi. Also cherry picked commit from #171 because that PR is having trouble merging.


adampbeardsley:
Really this branch is all about getting master to pass after dependency updates. Retitling the PR now.

PR #176, opened 2018-07-18
Pass tests

adampbeardsley:
Get rid of out-dated conda version specification, and fix bug so tests pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ install:
   - export PATH="$HOME/miniconda/bin:$PATH"
   - hash -r
   - conda config --set always_yes yes --set changeps1 no
-  - conda install -q conda=4.3.25
+  - conda install -q conda
   # Useful for debugging any issues with conda
   - conda info -a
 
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -17,7 +17,7 @@ install:
   - export PATH="$HOME/miniconda/bin:$PATH"
   - hash -r
   - conda config --set always_yes yes --set changeps1 no
-  - conda install -q conda=4.3.25
+  - conda install -q conda
```
mkolopanis: Was this specific conda version limiting?

hera_qm/omnical_metrics.py
```diff
@@ -802,9 +802,10 @@ def plot_chisq_tavg(self, pol_index=0, ants=None, ax=None, save=False, fname=Non
                 fname = os.path.join(self.filedir, fname)
             else:
                 fname = os.path.join(outpath, fname)
-        if custom_ax == False:
+        print('custom ax is : ', custom_ax)
```
mkolopanis: Is this a debug line? I really don't care if it stays just wonder if it is worth it.

PR #180, opened 2018-08-07
Make ant metrics faster

ianmalcolm:
Changes:
De-duplicate the redundant calculations in mean_Vij_metrics
Replace pyuvdata.uvdata.UVData with hera_cal.io.HERAData

Pros:
1.58x Faster
Cons:
Miss the support for fhd format


mkolopanis:
Looks like the tests are failing for more than just what was expected for the pyuvdata integration_time update.

 the `AttributeError: fake_data instance has no attribute '__getitem__'` seems be showing up a lot. Can you delve into this? just by looking at the definition it has this functionality, did something change?

PR #181, opened 2018-08-07
Integration time fix

mkolopanis:
This will still fail travis until the pyuvdata issue with integration_time cast as float32 is addressed. 

Fixed an issue where np.random.seed was not set in a test_xrfi function and test failed.

checks if integration_time is the same for all times in vis_count. if true just grab the first time, currently
raise NotImplementedError if the integration time varies with time.

On 3rd thought the line that checks this looks like len(set(integration_time[inds])), this could still fail if there are floating point errors, perhaps it should get changed to something like an all close

### DIFF
--- a/hera_qm/vis_metrics.py
+++ b/hera_qm/vis_metrics.py
@@ -24,7 +24,6 @@ def check_noise_variance(data):
         inds = data.antpair2ind(key[0], key[1])
         integration_time = data.integration_time[inds]
         if not len(set(integration_time)) == 1:
-            # what do if not same?
             raise NotImplementedError(("Integration times which vary with "
                                        "time are currently not supported."))
         else:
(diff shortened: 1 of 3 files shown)


hera_qm/vis_metrics.py
```diff
@@ -6,26 +6,41 @@
...
-    '''
+        Cij (dict): dictionary of variance measurements
+                    has keywords of (ant1, ant2, pol)
+    """
     Cij = {}
     for key, d in data.antpairpol_iter():
+        inds = data.antpair2ind(key[0], key[1])
+        integration_time = data.integration_time[inds]
+        if not len(set(integration_time)) == 1:
+            # what do if not same?
```
adampbeardsley: I don't think this comment is needed.

PR #182, opened 2018-08-09
.travis pyuvdata install commit 9c94be0366eb41692f

nkern:
Sets `pyuvdata` installation to commit 9c94be0366eb41692f (HEAD on master currently 8/9/2018)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ install:
   - conda install scikit-learn
   - conda install -c conda-forge aipy
   - pip install coveralls
-  - pip install git+https://github.com/HERA-Team/pyuvdata.git
+  - pip install git+https://github.com/HERA-Team/pyuvdata.git@9c94be0366eb41692f5501caac99981026a7ffe6
   - pip install git+https://github.com/HERA-Team/omnical.git
   - pip install git+https://github.com/HERA-Team/linsolve.git
   - pip install git+https://github.com/HERA-Team/uvtools.git


mkolopanis:
Dependabot sounds like it is worth looking into.

adampbeardsley:
So I just installed dependabot... and I still have no idea how it works. I'll poke around for a bit.

adampbeardsley:
Ok, from what I can tell dependabot reads a requirements.txt file to determine if our dependancies need updating.
https://pip.readthedocs.io/en/1.1/requirements.html
I think what we can do is make that requirements file, then have travis use that file to install things, so our tests will automatically run on the updated packages when dependabot submits a PR. I'm not sure if it will render specific commits though... we may have to depend on version/subversions. We could try it will the current hash and see if it makes a PR next time pyuvdata gets updated. Or we could test it on a branch with a purposely out of date hash. Anyone have a preference?

mkolopanis:
I have none. The more variables you control with it though the better we'll understand what works and what doesn't though right?

adampbeardsley:
Ok, I have some good news and some bad news. The good news is dependabot seems to do a thing (see PR #184 on branch `get_dependabot_to_work`). The bad news is it didn't play well with specific hashes - it required a version that pip knows about.
With pyuvdata pushing towards more frequent version bumps, this might be a viable solution. But we also probably want a patch solution for cases when we can't wait for pypi...

nkern:
My stance is that b/c software development for analysis routines fairly often involves `pyuvdata` contributions, it would be good to build against `pyuvdata` commits rather than releases: that way we can bump the commit up whenever we want and continue `hera_qm` development w/o having to wait for `pyuvdata` to make a new release. 

If we'd like to get feedback on API changes etc. w/ the HEAD of `pyuvdata` master, one thing we can do is make two build environments--one for a stable commit and one for master `HEAD`--and only make checks on the stable commit (something @philbull just implemented for `hera_pspec`).

What do people think about this?

adampbeardsley:
I like that idea @nkern, but is there a way to allow merging if tests pass on one environment (stable commit), but fail on the other (HEAD)? I don't see that kind of granularity, so I think it would effectively be the same as building only on HEAD.

adampbeardsley:
After some discussion on the telecon today, we agreed that an approach similar to hera_cal works best - we by default should be building against HEAD, but if you're in the middle of a PR and things start to fail due to a pyuvdata update, we can then specify a commit like this, and open an issue to bring things back up to HEAD.
@nkern does that plan sound ok to you? It would mean not merging this PR since it is currently pointing at head anyway.

nkern:
@adampbeardsley @jsdillon sorry I missed that telecon--I think we should probably do the inverse of that: we should by default build against something stable that *isn't changing*, like a commit, and

plaplant:
I don't want to speak for the others, but my opinion/recollection is that we wanted to stay on the bleeding edge as much as possible, because otherwise we'd have to remember to make a periodic PR just to increment the pyuvdata version being built against. Frankly I think we might forget to do that regularly, or have significant time go by where we may not be compatible with the latest version of master. By always trying to build against the current master, if something breaks and you have to peg to a specific version to usher the PR through, you can simultaneously make an issue to harmonize the repo with the latest changes when your PR is merged in, to make sure that we get back to master quickly and without introducing unrelated changes to the PR at hand.

To answer the point raised about figuring out which "last best version" should be pegged to when tests start to fail, we can (and probably should regardless) add a line to the travis build where we print the git hash of pyuvdata once it's installed. That way we can always go into the travis logs to find the hash for the last successful build on a branch; as a fallback, we can use the hash for the most recent master build. I understand the desire to introduce changes in a predictable rather than ad hoc fashion to avoid the issue mid-PR, but I think we would build up less technical debt if we only peg when necessary.

jsdillon:
I agree with Paul

adampbeardsley:
Over the course of a few telecons we've agreed that sticking with the bleeding edge by default is the way to go. If we need to build against a specific version to merge a PR, we can implement something like this, but immediately file and issue to get things up to speed.

PR #183, opened 2018-08-10
Noise modeling

nkern:
adds a sequential differencer to get noise models from visibility data

### DIFF
--- a/hera_qm/utils.py
+++ b/hera_qm/utils.py
@@ -462,7 +462,8 @@ def dynamic_slice(arr, slice_obj, axis=-1):
     """
     if isinstance(axis, (int, np.integer)):
         axis = (axis,)
-    assert isinstance(arr, np.ndarray), "arr must be an ndarray"
+    if not isinstance(arr, np.ndarray):
+        raise ValueError("arr must be an ndarray")
     slices = [slice(None) for i in range(arr.ndim)]
     for ax in axis:
         slices[ax] = slice_obj
--- a/hera_qm/utils.py
+++ b/hera_qm/utils.py
@@ -467,4 +467,4 @@ def dynamic_slice(arr, slice_obj, axis=-1):
     slices = [slice(None) for i in range(arr.ndim)]
     for ax in axis:
         slices[ax] = slice_obj
-    return arr[slices]
+    return arr[tuple(slices)]
--- a/hera_qm/utils.py
+++ b/hera_qm/utils.py
@@ -442,3 +442,28 @@ def metrics2mc(filename, ftype):
         raise ValueError('Metric file type ' + ftype + ' is not recognized.')
 
     return d
+
+
+def dynamic_slice(arr, slice_obj, axis=-1):
+    """
+    Dynamic slicing of arr along axis given slice object.
+
+    Parameters
+    ----------
+    arr : ndarray
+
+    slice_obj : slice object
+
+    axis : integer
+
+    Returns
+    -------
+    arr_slice : ndarray
+    """
+    if isinstance(axis, (int, np.integer)):
+        axis = (axis,)
+    assert isinstance(arr, np.ndarray), "arr must be an ndarray"
+    slices = [slice(None) for i in range(arr.ndim)]
+    for ax in axis:
+        slices[ax] = slice_obj
+    return arr[slices]
--- a/hera_qm/vis_metrics.py
+++ b/hera_qm/vis_metrics.py
@@ -125,7 +125,7 @@ def sequential_diff(data, wgts=None, axis=(0,)):
         bls = uvd.get_antpairs()
         for bl in bls:
             # get blt slice
-            bl_slice = uvd.get_blt_inds(bl)
+            bl_slice = uvd.antpair2ind(bl, ordered=False)
 
             # configure data and weights
             d = data.get_data(bl, squeeze='none')[:, 0, :, :]
--- a/hera_qm/vis_metrics.py
+++ b/hera_qm/vis_metrics.py
@@ -50,9 +50,9 @@ def check_noise_variance(data):
     return Cij
 
 
-def sequential_diff(data, wgts=None, axis=(0,)):
+def sequential_diff(data, t_int=None, axis=(0,), pad=False):
     """
-    Take a sequential (adjacent) difference of a visibility waterfall
+    Take a sequential (forward) difference of a visibility waterfall
     as an estimate of the visibility noise.
 
     Parameters
@@ -62,13 +62,18 @@ def sequential_diff(data, wgts=None, axis=(0,)):
         shape (Ntimes, Nfreqs, :). Or UVData object
         holding visibility data.
 
-    wgts : ndarray
-        A 2D or 3D ndarray containing visibility weights
-        matching data shape. If None, and data is UVData,
-        will use UVData ~flags * nsample as weights.
+    t_int : ndarray
+        A 2D or 3D ndarray containing the integration time
+        of the visibilities matching shape of data.
+        If None and data is UVData, will use UVData ~flags *
+        nsample * integration_time as input.
 
     axis : int or tuple
         Axes along which to take sequential difference
+
+    pad : boolean
+        If True, insert an extra (flagged) column at the end
+        of axis such that diff_data has same shape as input.
 
     Returns
     -------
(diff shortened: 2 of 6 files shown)


nkern:
Any idea what this error is coming from
```
Traceback (most recent call last):
  File "/home/travis/build/HERA-Team/hera_qm/hera_qm/tests/test_ant_metrics.py", line 296, in test_cross_detection
    fileformat='miriad')
  File "/home/travis/build/HERA-Team/hera_qm/hera_qm/ant_metrics.py", line 325, in __init__
    self.data.read_miriad(dataFileList)
  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/pyuvdata/uvdata.py", line 1865, in read_miriad
    polarizations=polarizations, time_range=time_range)
  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/pyuvdata/uvdata.py", line 1887, in read_miriad
    polarizations=polarizations, time_range=time_range)
  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/pyuvdata/miriad.py", line 381, in read_miriad
    self.set_lsts_from_time_array()
  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/pyuvdata/uvdata.py", line 562, in set_lsts_from_time_array
    self.lst_array = uvutils.get_lst_for_time(self.time_array, latitude, longitude, altitude)
  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/pyuvdata/utils.py", line 432, in get_lst_for_time
    jd, jd_array, atol=1e-6, rtol=1e-12))] = t.sidereal_time('apparent').radian
  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/astropy/time/core.py", line 758, in sidereal_time
    gst = self._erfa_sidereal_time(available_models[model.upper()])
  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/astropy/time/core.py", line 774, in _erfa_sidereal_time
    for jd_part in ('jd1', 'jd2')]
  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/astropy/time/core.py", line 1125, in __getattr__
    tm._set_scale(attr)
  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/astropy/time/core.py", line 496, in _set_scale
    args.append(get_dt(jd1, jd2))
  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/astropy/time/core.py", line 1247, in _get_delta_ut1_utc
    delta = iers_table.ut1_utc(jd1, jd2)
  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/astropy/utils/iers/iers.py", line 229, in ut1_utc
    self.ut1_utc_source if return_status else None)
  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/astropy/utils/iers/iers.py", line 327, in _interpolate
    self._check_interpolate_indices(i1, i, np.max(mjd))
  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/astropy/utils/iers/iers.py", line 270, in _check_interpolate_indices
    raise IERSRangeError('(some) times are outside of range covered '
IERSRangeError: (some) times are outside of range covered by IERS table.
```

adampbeardsley:
That error typically happens when the navy server is down. I restarted your build, hopefully it will work this time.

adampbeardsley:
Ok something very confusing is going on. I've restart your build a few times, and it keeps timing out in different ways. Let's maybe be patient and see if it resolves itself eventually?

hera_qm/vis_metrics.py
```diff
@@ -48,6 +50,105 @@ def check_noise_variance(data):
...
+    assert np.all([ax in [0, 1] for ax in axis]), "axis must be 0, 1 or both"
+
+    # perform differencing if data is an ndarray
+    if isinstance(data, np.ndarray):
+        # get weights
+        if wgts is None:
+            wgts = np.ones_like(data, dtype=np.float64)
+        else:
+            assert isinstance(wgts, np.ndarray), \
+                   "wgts must be ndarray if data is ndarray"
```
adampbeardsley: Please us if/else instead of assert.

hera_qm/utils.py
```diff
@@ -442,3 +442,28 @@ def metrics2mc(filename, ftype):
...
+
+    axis : integer
+
+    Returns
+    -------
+    arr_slice : ndarray
+    """
+    if isinstance(axis, (int, np.integer)):
+        axis = (axis,)
+    assert isinstance(arr, np.ndarray), "arr must be an ndarray"
```
adampbeardsley: Could you put this in an if statement instead? Paul has pointed out you can run code in a way that ignore asserts, so we're generally avoiding them in hera_qm.

hera_qm/vis_metrics.py
```diff
@@ -48,6 +50,105 @@ def check_noise_variance(data):
...
+        # get new time and freq arrays and UVData object
+        times = None
+        freqs = None
+        for ax in axis:
+            if ax == 0:
+                times = np.unique(uvd.time_array)[:-1]
+            elif ax == 1:
+                freqs = np.unique(uvd.freq_array)[:-1]
+        if times is not None or freqs is not None:
+            uvd.select(times=times, frequencies=freqs)
```
adampbeardsley: Shouldn't the freqs and times be the average of the bins going into the subtraction?

hera_qm/vis_metrics.py
```diff
@@ -48,6 +50,105 @@ def check_noise_variance(data):
...
+        else:
+            assert isinstance(wgts, np.ndarray), \
+                   "wgts must be ndarray if data is ndarray"
+
+        # difference data
+        for ax in axis:
+            data = (utils.dynamic_slice(data, slice(1, None), axis=ax) \
+                   - utils.dynamic_slice(data, slice(None, -1), axis=ax)) / np.sqrt(2)
+            wgts = np.sqrt(utils.dynamic_slice(wgts, slice(1, None), axis=ax) \
+                   * utils.dynamic_slice(wgts, slice(None, -1), axis=ax))
```
adampbeardsley: Why is it that weights are multiplied? Eventually this gets reinterpreted as flags and nsamples. I can see how this gets you the correct flagging (if any component is flagged, the result is flagged), but it doesn't seem like the right thing for nsamples.

hera_qm/vis_metrics.py
nkern: I think it could be either one depending on your interpretation: sort of like the different finite-differencing scheme.. One thing I don't like about changing the times and frequencies is that now they are mismatched w.r.t. the data we started with, which if possible I'd like to try to retain.

hera_qm/vis_metrics.py
nkern: yeah it was done to get the correct flagging patterns, but you are right that its not strictly correct for `nsample`. In fact, upon reflection, I think `wgts` is a bad name b/c we aren't doing any kind of weighting anyways. I'll change this

nkern:
I think this is ready to go, but before we finish I want to see if people have input as to why and how I'm updating the `nsample` array after taking the finite difference. The idea is that after every finite difference, I'm boosting the noise by a factor of sqrt(2), so I divide the nsample_array by sqrt(2) every finite difference I do. However, I'm also dividing the data itself by sqrt(2) to not bias the end-result noise estimate of the initial data I started with. Am I double counting the bias correction? Perhaps I should either 1. correct the data and not the nsample array or 2. correct the nsample array and not the data.

What do you think @adampbeardsley ?

nkern:
I think what I've convinced myself is that the correct solution is to update the `nsample_array`, and let the user properly normalize the noise amplitude after-the-fact. The advantage here being that the way I update the `nsample_array` can account for differencing two pixels w/ different inherent integration time, which would be wrong if I just applied a correction of `1/sqrt(2)` to the data after differencing.

hera_qm/utils.py
```diff
@@ -442,3 +442,29 @@ def metrics2mc(filename, ftype):
...
+    arr_slice : ndarray
+    """
+    if isinstance(axis, (int, np.integer)):
+        axis = (axis,)
+    if not isinstance(arr, np.ndarray):
+        raise ValueError("arr must be an ndarray")
+    slices = [slice(None) for i in range(arr.ndim)]
+    for ax in axis:
+        slices[ax] = slice_obj
+    return arr[slices]
```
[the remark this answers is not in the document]
nkern: I can't reproduce this warning. What python version and how are you calling it?

hera_qm/vis_metrics.py
```diff
@@ -48,6 +51,138 @@ def check_noise_variance(data):
...
+                # pad arrays
+                data = np.concatenate([data, p], axis=ax)
+                t_int = np.concatenate([t_int, p], axis=ax)
+
+        # difference data
+        for ax in axis:
+            data = (utils.dynamic_slice(data, slice(1, None), axis=ax) \
+                   - utils.dynamic_slice(data, slice(None, -1), axis=ax))
+            t_int = 1.0/(1.0/utils.dynamic_slice(t_int, slice(1, None), axis=ax) \
+                         + 1.0/utils.dynamic_slice(t_int, slice(None, -1), axis=ax))
```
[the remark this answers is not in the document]
nkern: I don't have a problem w/ dividebyzero warning here, b/c that is indeed what I want it to do: the 1./0. will give me an inf and the 1/inf will then give me a zero, which is what I want.

If I use the `true_divide` then the zeros are left untouched, and that means when I do the sum against another array which might not be zero, then it won't do what I want.

For example, if
```
arr1 = [1.0, 1.0, 0.0]
arr2 = [1.0, 1.0, 1.0]
```
I want the inverse average of these to be
```
inv_avg = [0.5, 0.5, 0.0]
```
But in this case if I just do
```
s = np.true_divide(1, arr1, where=~np.isclose(arr1, 0.0)) + np.true_divide(1, arr2, where=~np.isclose(arr2, 0.0))
t_int = np.true_divide(1, s, where=~np.isclose(s, 0.0))
```
then I end up with
```
[0.5, 0.5, 1. ]
```

hera_qm/utils.py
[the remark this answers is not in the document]
adampbeardsley: I'm running python 2.7.15 and numpy 1.15.0. All I did to get the warning was called your function like so:
`dynamic_slice(np.zeros((5,5)),slice(None))`

hera_qm/vis_metrics.py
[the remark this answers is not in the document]
adampbeardsley: Let's work on this. Using divide by zeros can be dangerous, and because of how python reports warnings, it could potentially hide other warnings elsewhere (each warning is only reported once). Given what you wrote above, I think the following will do what you want:
```s = np.true_divide(1, arr1, where=~np.isclose(arr1, 0.0)) + np.true_divide(1, arr2, where=~np.isclose(arr2, 0.0))
t_int = np.true_divide(1, s, where=~np.isclose(s, 0.0))
t_int = np.where(0, t_int, np.isclose(s, 0.0))
```
@plaplant I know you've dealt with stuff like this before, do you have a more elegant idea?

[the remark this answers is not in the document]
plaplant: @adampbeardsley I 100% agree that we should avoid dividing by zero as much as possible. I don't have a more elegant solution than that, but I think `np.where` is relatively cheap, so it shouldn't add too much overhead.

hera_qm/utils.py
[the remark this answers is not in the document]
nkern: Ah okay, I see. thanks

nkern:
@adampbeardsley okay, I think this is ready to go then. let me know if you have any further comments

nkern:
So now there is a weird error coming from `aipy` that I can't reproduce locally from an older version of `aipy`:
```
======================================================================
ERROR: hera_qm.tests.test_ant_metrics.TestAntmetricsRun.test_ant_metrics_run
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/build/HERA-Team/hera_qm/hera_qm/tests/test_ant_metrics.py", line 362, in test_ant_metrics_run
    ant_metrics.ant_metrics_run(args.files, args, history)
  File "/home/travis/build/HERA-Team/hera_qm/hera_qm/ant_metrics.py", line 555, in ant_metrics_run
    aa = aipy.cal.get_aa(args.cal, freqs)
  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/aipy/cal.py", line 30, in get_aa
    return _get_aa(freqs)
  File "/home/travis/build/HERA-Team/hera_qm/hera_qm/data/heratest_calfile.py", line 274, in get_aa
    beam.set_params(prms['bm_prms'])
  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/aipy/fit.py", line 291, in set_params
    data.shape = (data.size/2, 2)
TypeError: 'float' object cannot be interpreted as an index
```
@pkgw anything from a recent release of `aipy` that would cause this?

hera_qm/vis_metrics.py
```diff
@@ -48,6 +51,156 @@ def check_noise_variance(data):
...
+        nsample * integration_time as input.
+
+    axis : int or tuple
+        Axes along which to take sequential difference
+
+    pad : boolean
+        If True, insert an extra (flagged) column at the end
+        of axis such that diff_data has same shape as input.
+
+    history : str
```
adampbeardsley: I just noticed while reviewing your latest commits, that this argument isn't actually used in the function.

adampbeardsley:
Hmm, I think that error looks legit, at least since aipy because py3 compatible, and numpy started requiring that indices be integers. We'll have to dig into that code to try and fix it. I can try to take a look monday if people don't get to it sooner.

pkgw:
This should be fixed in aipy version 3.0.0rc2, which will be hitting pip and conda-forge soon.

hera_qm/vis_metrics.py
nkern: ah thanks, fixed that

PR #195, opened 2018-08-27
Makes repo Python3 compatible

nkern:


mkolopanis:
This PR will need to rebase (or merge in) master  to get up to date after both revamps have been merged in.

hera_qm/tests/test_xrfi.py
```diff
@@ -1,8 +1,8 @@
 # -*- coding: utf-8 -*-
 # Copyright (c) 2018 the HERA Project
 # Licensed under the MIT License
-
 from __future__ import print_function, division, absolute_import
+import hera_qm.tests as qmtest
```
adampbeardsley: err, this import already shows up below (line 14). Also, it's not used. So let's chop this line and 14.

hera_qm/firstcal_metrics.py
```diff
@@ -520,14 +524,20 @@ def write_metrics(self, filename=None, filetype='json'):
...
+                if isinstance(metrics_out[k], np.ndarray):
+                    metrics_out[k] = metrics_out[k].tolist()
+                elif isinstance(metrics_out[k], np.float):
+                    metrics_out[k] = float(metrics_out[k])
+                elif isinstance(metrics_out[k], np.integer):
+                    metrics_out[k] = int(metrics_out[k])
+                elif isinstance(metrics_out[k], np.bool):
+                    metrics_out[k] = bool(metrics_out[k])
+                elif isinstance(metrics_out[k], (dict, odict)):
+                    metrics_out[k] = odict([(str(_k), metrics_out[k][_k].tolist()) if isinstance(metrics_out[k][_k], np.ndarray) else (str(_k), metrics_out[k][_k]) for _k in metrics_out[k]])
```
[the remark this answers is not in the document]
mkolopanis: I think this line is at least worth break so the if/else happens on separate lines, but eventually this should all be taken handled with `metrics_io` so I'm not sure it is worth digging in too deep here.

hera_qm/tests/test_xrfi.py
mkolopanis: image

PR #202, opened 2018-09-11
Read old ant_metrics json files

mkolopanis:
added new parsing type for old ant_metrics files and test file.

checks if an expected dict of dict of dicts is actually a list and converts the old ant_metrics jsons list of dict of dicts to a dict of dict of dicts as it is read in.


hera_qm/metrics_io.py
```diff
@@ -564,14 +595,21 @@ def _parse_dict_of_dict_of_dicts(input_str, value_type=float):
...
+
     dict_regex = r"([0-9]*?)\'?: \{(.*?\})\}"
     dicts = re.findall(dict_regex, input_str)
 
     # initialize output
     output = {}
     for d in dicts:
         # key is first capture group
-        key = int(str(d[0]))
+        # key = int(str(d[0]))
```
plaplant: This should be removed instead of commented out.

hera_qm/tests/test_metrics_io.py
```diff
@@ -380,3 +380,36 @@ def test_read_write_new_ant_json_files():
...
+    # so ignore them.
+    test_metrics_old.pop('history', None)
+    test_metrics_old.pop('version', None)
+    test_metrics_new.pop('history', None)
+    test_metrics_new.pop('version', None)
+
+    # This function recursively walks dictionary and compares
+    # data types together with nt.assert_type_equal or np.allclose
+    qmtest.recursive_compare_dicts(test_metrics_new, test_metrics_old)
+    # nt.assert_true(os.path.exists(test_file))
```
plaplant: This line and the next should either not be commented out (preferred) or removed.

hera_qm/metrics_io.py
mkolopanis: oh good catch thanks. I commented it to test using the internal parser and never removed

hera_qm/tests/test_metrics_io.py
mkolopanis: Not sure why I commented them to begin with.

PR #206, opened 2018-09-16
Process exants

### LINKED ISSUE #205: Acquire process_exants from hera_cal
hera_cal will be deprecating `process_exants`, but we still use it in xrfi. We need to either take ownership of it, or write it out of our code.

adampbeardsley:
This PR steals the process_ex_ants function from hera_cal. In the process, I slightly changed the interface so the two ways of passing exants in are both optional keywords now.

fixes #205

### DIFF
--- a/hera_qm/tests/test_utils.py
+++ b/hera_qm/tests/test_utils.py
@@ -404,21 +404,3 @@ def test_dynamic_slice():
     nt.assert_equal(b.shape, (1, 5))
 
     nt.assert_raises(ValueError, utils.dynamic_slice, 'foo', slice(0, None))
-
-def test_process_ex_ants():
-    ex_ants = ''
-    xants = utils.process_ex_ants(ex_ants=ex_ants)
-    nt.assert_equal(xants, [])
-
-    ex_ants = '0,1,2'
-    xants = utils.process_ex_ants(ex_ants=ex_ants)
-    nt.assert_equal(xants, [0, 1, 2])
-
-    ex_ants = '0,obvious_error'
-    nt.assert_raises(AssertionError, utils.process_ex_ants, ex_ants=ex_ants)
-
-    ex_ants = '0,1'
-    met_file = os.path.join(DATA_PATH, 'example_ant_metrics.hdf5')
-    xants = utils.process_ex_ants(ex_ants=ex_ants, metrics_file=met_file)
-    nt.assert_equal(xants, [0, 1, 81])
-    return
--- a/hera_qm/tests/test_utils.py
+++ b/hera_qm/tests/test_utils.py
@@ -404,3 +404,21 @@ def test_dynamic_slice():
     nt.assert_equal(b.shape, (1, 5))
 
     nt.assert_raises(ValueError, utils.dynamic_slice, 'foo', slice(0, None))
+
+def test_process_ex_ants():
+    ex_ants = ''
+    xants = omni.process_ex_ants(ex_ants)
+    nt.assert_equal(xants, [])
+
+    ex_ants = '0,1,2'
+    xants = omni.process_ex_ants(ex_ants)
+    nt.assert_equal(xants, [0, 1, 2])
+
+    ex_ants = '0,obvious_error'
+    nt.assert_raises(AssertionError, omni.process_ex_ants, ex_ants)
+
+    ex_ants = '0,1'
+    met_file = os.path.join(DATA_PATH, 'example_ant_metrics.hdf5')
+    xants = omni.process_ex_ants(ex_ants, metrics_json=json_file)
+    nt.assert_equal(xants, [81])
+    return
--- a/hera_qm/tests/test_utils.py
+++ b/hera_qm/tests/test_utils.py
@@ -407,18 +407,18 @@ def test_dynamic_slice():
 
 def test_process_ex_ants():
     ex_ants = ''
-    xants = omni.process_ex_ants(ex_ants)
+    xants = utils.process_ex_ants(ex_ants=ex_ants)
     nt.assert_equal(xants, [])
 
     ex_ants = '0,1,2'
-    xants = omni.process_ex_ants(ex_ants)
+    xants = utils.process_ex_ants(ex_ants=ex_ants)
     nt.assert_equal(xants, [0, 1, 2])
 
     ex_ants = '0,obvious_error'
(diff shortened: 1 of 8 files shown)


hera_qm/tests/test_utils.py
```diff
@@ -404,3 +404,21 @@ def test_dynamic_slice():
     nt.assert_equal(b.shape, (1, 5))
 
     nt.assert_raises(ValueError, utils.dynamic_slice, 'foo', slice(0, None))
+
+def test_process_ex_ants():
```
mkolopanis: This one mega test looks like it is testing the different returns and failure routes of process_ex_ants. It would be a little more helpful from a debugging standpoint if each `assert`-like statement was its own test. Then we can pinpoint issues in futures updates if there are any. I know there are a *LOT* of tests like this in the current version of `hera_qm` and a lot of it is legacy code but if we get better about writing new tests and code better there will be less technical debt later. My two cents.

hera_qm/tests/test_xrfi.py
```diff
@@ -624,20 +624,19 @@ def test_xrfi_h1c_run(self):
...
-                             'Just as test.', filename=test_d_file + '.h1c_run')
-
-            # test no filename provided
-            nt.assert_raises(AssertionError, xrfi.xrfi_h1c_run, uvd,
-                             'Just as test.', filename=None)
+        uvtest.checkWarnings(xrfi.xrfi_h1c_run, [None],
+                             {'filename': test_d_file, 'history': 'Just a test.',
+                              'model_file': test_d_file, 'model_file_format': 'miriad',
+                              'xrfi_path': xrfi_path}, nwarnings=191,
+                              message=['indata is None'] + 190 * ['Kt value 8'])
```
mkolopanis: a very specific number of Kt value 8 warnings :open_mouth:

hera_qm/tests/test_xrfi.py
```diff
@@ -624,20 +624,19 @@ def test_xrfi_h1c_run(self):
...
+
+        # test no indata provided
+        nt.assert_raises(AssertionError, xrfi.xrfi_h1c_run, None,
```
mkolopanis: See my previous rant about testing etiquette.

hera_qm/utils.py
```diff
@@ -659,3 +659,38 @@ def dynamic_slice(arr, slice_obj, axis=-1):
...
+def process_ex_ants(ex_ants=None, metrics_file=None):
+    """
+    Return list of excluded antennas from command line argument.
+    Input:
+       ex_ants -- comma-separated value list of excluded antennas
+       metrics_file -- file containing array info from hera_qm
+    Output:
+       list of excluded antennas
+    """
+    from . import metrics_io
```
mkolopanis: This is to circumvent complete circular dependence between utils and metrics_io right? @plaplant  is there a better way (other than this function having its own module)?

hera_qm/xrfi.py
```diff
@@ -10,6 +10,7 @@
 from .uvflag import UVFlag
 from . import utils as qm_utils
 from .version import hera_qm_version_str
+from .utils import process_ex_ants
```
mkolopanis: do we need this import? is if possible to just call `qm_utils.process_ex_ants` since you already have utils imported? I guess it doesn't really matter either way to be honest.

hera_qm/utils.py
plaplant: Yeah, I don't think there's a way around burying the import here (which isn't to my personal coding taste) besides moving it to a different file. Could we just have `process_ex_ants` be in the `metrics_io` module? I know it's not a perfect fit, but it would remove the circular dependence issue.

hera_qm/utils.py
```diff
@@ -183,10 +183,10 @@ def get_metrics_ArgumentParser(method_name):
         a.add_argument('--time_threshold', default=0.05, type=float,
                        help='Fraction of times required to trigger broadcast across'
                        ' time (single frequency). Default is 0.05.')
-        a.add_argument('--ex_ants', default='', type=str,
+        a.add_argument('--ex_ants', default=None, type=str,
                        help='Comma-separated list of antennas to exclude. Flags of visibilities '
                        'formed with these antennas will be set to True.')
-        a.add_argument('--metrics_json', default='', type=str,
+        a.add_argument('--metrics_json', default=None, type=str,
```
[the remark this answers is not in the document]
adampbeardsley: yeah, I was torn about this. I thought we'd want to keep the old naming because this is a legacy wrapper as it is. But I'm happy to change it if I'm the minority on this.

hera_qm/tests/test_xrfi.py
adampbeardsley: ikr

hera_qm/xrfi.py
adampbeardsley: good catch. will change.

adampbeardsley: Actually, I like Paul's suggestion to put process_ex_ants in metrics_io, so that will supersede this.

hera_qm/utils.py
adampbeardsley: I like that solution best. I'll move it to metrics_io.

hera_qm/utils.py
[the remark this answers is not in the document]
mkolopanis: could always overload the destination with both inputs. have lines like:
``` 
a.add_argument('--metrics_file', dest='metrics_file', type=str, action='append')
a.add_argument('--metrics_json', dest='metrics_file', type=str, action='append')
a.set_defaults(metrics_file=None)
```
that way someone could specify either or both of the inputs. The draw back is that we could end up with something like this:
```
In [17]: a.parse_args('--metrics_file=foobar --metrics_json=foobar'.split())
Out[17]: Namespace(metrics_file=['foobar', 'foobar'])
```
but if we're not worried about getting the same argument multiple times it doesn't matter.  Especially since this is for `process_ex_ants`, it won't append the same antenna multiple times and if you loop over a file twice it won't take long and it was your fault anyway.

hera_qm/xrfi.py
mkolopanis: :+1:

hera_qm/utils.py
mkolopanis: :+1:

PR #210, opened 2018-09-18
Uvh5 xrfi

### LINKED ISSUE #209: Make xrfi uvh5 compatible

adampbeardsley:
Adds support for uvh5 files in xrfi. Actually, it removes the places where we specify the infile format and instead uses pyuvdata's generic read function. The one place where we write data, we can now use uvh5.
Fixes #209

### DIFF
--- a/hera_qm/utils.py
+++ b/hera_qm/utils.py
@@ -138,7 +138,7 @@ def get_metrics_ArgumentParser(method_name):
     elif method_name == 'xrfi_run':
         a.prog = 'xrfi_run.py'
         a.add_argument('--infile_format', default='miriad', type=str,
-                       help='File format for input files. Default is miriad.')
+                       help='File format for input files. DEPRECATED, but kept for legacy.')
         a.add_argument('--extension', default='.flags.npz', type=str,
                        help='Extension to be appended to input file name. Default is ".flags.npz".')
         a.add_argument('--summary', action='store_true', default=False,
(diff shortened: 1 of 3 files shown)


hera_qm/utils.py
```diff
@@ -138,7 +138,7 @@ def get_metrics_ArgumentParser(method_name):
     elif method_name == 'xrfi_run':
         a.prog = 'xrfi_run.py'
         a.add_argument('--infile_format', default='miriad', type=str,
-                       help='File format for input files. Default is miriad.')
+                       help='File format for input files. DEPRECATED, but kept for legacy.')
```
plaplant: Do we want to get fancy and actually have a `DeprecationWarning` displayed to the user if they supply this argument?

plaplant: Also, why is it deprecated? Because we use the generic `read` method in pyuvdata?

hera_qm/xrfi.py
```diff
@@ -669,12 +669,7 @@ def xrfi_h1c_run(indata, history, infile_format='miriad', extension='.flags.h5',
         elif not isinstance(filename, str):
             raise ValueError('filename must be string path to file.')
         uvd = UVData()
-        if infile_format == 'miriad':
-            uvd.read_miriad(filename)
-        elif infile_format == 'uvfits':
-            uvd.read_uvfits(filename)
-        else:
-            raise ValueError('Unrecognized input file format ' + str(infile_format))
+        uvd.read(filename)
```
plaplant: Ah I see we are using `read`, so that makes sense.

PR #211, opened 2018-09-27
break out collapse_pol function in uvflag

adampbeardsley:
This function allows the user to collapse the polarization dimension. Previously it was done in the `to_waterfall` function if `keep_pols == False`, but this will expose the functionality to the user.

### DIFF
--- a/hera_qm/tests/test_uvflag.py
+++ b/hera_qm/tests/test_uvflag.py
@@ -612,8 +612,6 @@ def test_collapse_pol():
     print(uvf2.polarization_array)
     uvf2.collapse_pol()
     nt.assert_true(len(uvf2.polarization_array) == 1)
-    print(uvf2.polarization_array)
-    print(uvf.polarization_array)
     nt.assert_true(uvf2.polarization_array[0] == ','.join(map(str, uvf.polarization_array)))
 
 
@@ -628,8 +626,6 @@ def test_collapse_pol_flag():
     print(uvf2.polarization_array)
     uvf2.collapse_pol()
     nt.assert_true(len(uvf2.polarization_array) == 1)
-    print(uvf2.polarization_array)
-    print(uvf.polarization_array)
     nt.assert_true(uvf2.polarization_array[0] == ','.join(map(str, uvf.polarization_array)))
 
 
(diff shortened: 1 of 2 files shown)


hera_qm/uvflag.py
```diff
@@ -468,6 +465,29 @@ def to_waterfall(self, method='quadmean', keep_pol=True):
...
+        Args:
+            method: How to collapse the dimension(s)
+        """
+        method = method.lower()
+        avg_f = qm_utils.averaging_dict[method]
+        if self.mode == 'flag':
+            darr = self.flag_array
+        else:
+            darr = self.metric_array
+        if len(self.polarization_array) > 1:
```
mkolopanis: I understand there is nothing to collapse if `len(self.polarization_array)==1` but is a silent failure like this desirable? Or should there be some kind of warning or error raised? I think an error sounds too serious, but maybe it can raise a warning saying that there is nothing to collapse? Just to let the user know there is nothing to do or that it's not a reasonable call to make?

hera_qm/tests/test_uvflag.py
```diff
@@ -602,6 +602,37 @@ def test_to_waterfall_bl_multi_pol():
...
+    uvf.weights_array = np.ones_like(uvf.weights_array)
+    uvf2 = uvf.copy()
+    uvf2.polarization_array[0] = -4
+    uvf.__add__(uvf2, inplace=True, axis='pol')  # Concatenate to form multi-pol object
+    uvf2 = uvf.copy()
+    print(uvf2.polarization_array)
+    uvf2.collapse_pol()
+    nt.assert_true(len(uvf2.polarization_array) == 1)
+    print(uvf2.polarization_array)
+    print(uvf.polarization_array)
```
mkolopanis: can these print statements be removed? Were they just debugg stuff?

hera_qm/tests/test_uvflag.py
```diff
@@ -602,6 +602,37 @@ def test_to_waterfall_bl_multi_pol():
...
```
mkolopanis: see previous print statement comment.

hera_qm/tests/test_uvflag.py
adampbeardsley: whoops

adampbeardsley:
Thanks @mkolopanis . I think I've fixed these, ready for your reviewing pleasure.

PR #221, opened 2018-10-21
Output file convention

### LINKED ISSUE #220: Adhere to new filename convention
Namely, we've agreed to stop keeping file extensions in the middle of filenames. For example, we currently make file like ...`uvh5.ant_metrics.hdf5`, which we would like to make ...`ant_metrics.hdf5`.

adampbeardsley:
Fixes #220 

Also brought a lot of code up to pep8 compliance

### DIFF
--- a/hera_qm/tests/test_utils.py
+++ b/hera_qm/tests/test_utils.py
@@ -412,8 +412,13 @@ def test_strip_extension():
     nt.assert_equal(root, path[:-4])
 
 
-def test_strip_extension_return_ext():
+def test_strip_extension_return_ext_basename():
     path = 'goo/foo.boo/hoo/woo.two'
     root, ext = utils.strip_extension(path, return_ext=True)
     nt.assert_equal(root, path[:-4])
+
+
+def test_strip_extension_return_ext_extension():
+    path = 'goo/foo.boo/hoo/woo.two'
+    root, ext = utils.strip_extension(path, return_ext=True)
     nt.assert_equal(ext, path[-3:])
--- a/hera_qm/tests/test_utils.py
+++ b/hera_qm/tests/test_utils.py
@@ -98,7 +98,7 @@ def test_get_metrics_ArgumentParser_xrfi_run():
     # First try defaults - test a few of them
     args = a.parse_args('')
     nt.assert_equal(args.infile_format, 'miriad')
-    nt.assert_equal(args.summary_ext, '.flag_summary.npz')
+    nt.assert_equal(args.summary_ext, 'flag_summary.h5')
     nt.assert_equal(args.algorithm, 'xrfi_simple')
     nt.assert_equal(args.nsig_df, 6.0)
     nt.assert_equal(args.px_threshold, 0.2)
@@ -404,3 +404,16 @@ def test_dynamic_slice():
     nt.assert_equal(b.shape, (1, 5))
 
     nt.assert_raises(ValueError, utils.dynamic_slice, 'foo', slice(0, None))
+
+
+def test_strip_extension():
+    path = 'goo/foo.boo/hoo/woo.two'
+    root = utils.strip_extension(path)
+    nt.assert_equal(root, path[:-4])
+
+
+def test_strip_extension_return_ext():
+    path = 'goo/foo.boo/hoo/woo.two'
+    root, ext = utils.strip_extension(path, return_ext=True)
+    nt.assert_equal(root, path[:-4])
+    nt.assert_equal(ext, path[-3:])
--- a/hera_qm/tests/test_vis_metrics.py
+++ b/hera_qm/tests/test_vis_metrics.py
@@ -67,7 +67,7 @@ def test_vis_bl_cov():
     cov = vis_metrics.vis_bl_bl_cov(uvd, uvd, bls)
     nt.assert_equal(cov.shape, (6, 6, 1, 1))
     nt.assert_equal(cov.dtype, np.complex128)
-    nt.assert_almost_equal(cov[0,0,0,0], (51.06967733738634+0j))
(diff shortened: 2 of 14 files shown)


jsdillon:
Do you know about os.path.splitext() and os.path.basename()? That might get
you what you want without new functions.

hera_qm/tests/test_vis_metrics.py
```diff
@@ -185,9 +185,12 @@ def test_sequential_diff():
     uvn_diff3 = vis_metrics.sequential_diff(uvn, axis=(0, 1), pad=False)
 
     # assert noise std is equal to 1 within sampling error
-    nt.assert_almost_equal(np.std(uvn_diff1.data_array), 1.0, delta=1/np.sqrt(uvn.Ntimes * uvn.Nfreqs))
-    nt.assert_almost_equal(np.std(uvn_diff2.data_array), 1.0, delta=1/np.sqrt(uvn.Ntimes * uvn.Nfreqs))
-    nt.assert_almost_equal(np.std(uvn_diff3.data_array), 1.0, delta=1/np.sqrt(uvn.Ntimes * uvn.Nfreqs))
+    nt.assert_almost_equal(np.std(uvn_diff1.data_array), 1.0, delta=1
+                           / np.sqrt(uvn.Ntimes * uvn.Nfreqs))
```
mkolopanis: Maybe it is just me but it feels weird splitting the line in the middle of the definition of delta. Could we move the entire delta expression to its own line?

hera_qm/tests/test_utils.py
```diff
@@ -404,3 +404,16 @@ def test_dynamic_slice():
...
     nt.assert_raises(ValueError, utils.dynamic_slice, 'foo', slice(0, None))
+
+
+def test_strip_extension():
+    path = 'goo/foo.boo/hoo/woo.two'
+    root = utils.strip_extension(path)
+    nt.assert_equal(root, path[:-4])
+
+
+def test_strip_extension_return_ext():
```
mkolopanis: This suggestion will make a lot more sense when you look at the following comment. It's about being pedantic with exactly what you're testing. But not really necessary either, feel free to reject.
```suggestion
def test_strip_extension_return_ext_basename()
```

hera_qm/tests/test_utils.py
```diff
@@ -404,3 +404,16 @@ def test_dynamic_slice():
...
+    path = 'goo/foo.boo/hoo/woo.two'
+    root, ext = utils.strip_extension(path, return_ext=True)
+    nt.assert_equal(root, path[:-4])
```
mkolopanis: This assert is _technically_ redundant since it is executed in the previous test and all this test is supposed to do is test the extension returned by the helper function is correct. Or if you want to be really pedantic about if you could do this and rename the test to `test_strip_extension_return_ext_basename`
```suggestion
    nt.assert_equal(root, path[:-4])

def test_strip_extension_return_ext_extension():
    path = 'goo/foo.boo/hoo/woo.two'
    root, ext = utils.strip_extension(path, return_ext=True)
```

hera_qm/tests/test_vis_metrics.py
```diff
@@ -185,9 +185,12 @@ def test_sequential_diff():
...
+    nt.assert_almost_equal(np.std(uvn_diff2.data_array), 1.0, delta=1
+                           / np.sqrt(uvn.Ntimes * uvn.Nfreqs))
```
mkolopanis: see above

hera_qm/tests/test_vis_metrics.py
```diff
@@ -185,9 +185,12 @@ def test_sequential_diff():
...
+    nt.assert_almost_equal(np.std(uvn_diff3.data_array), 1.0, delta=1
+                           / np.sqrt(uvn.Ntimes * uvn.Nfreqs))
```
mkolopanis: see above

plaplant:
@adampbeardsley do we want to have `pycodestyle` run on the repo as part of PRs, like in pyuvdata, so that we don't get out of sync and have clean-up PRs like this one? I know sometimes it's annoying when the build fails for style reasons, but it seems like the better option if we want to be pep8 compliant.

hera_qm/tests/test_xrfi.py
```diff
@@ -725,9 +726,10 @@ def test_xrfi_h1c_run_input_calfits(self):
         # input calfits
         uvd = UVData()
         uvd.read_miriad(test_d_file)
-        ext = '.flag'
-        outtest1 = os.path.join(xrfi_path, os.path.basename(test_c_file)) + '.x' + ext
-        outtest2 = os.path.join(xrfi_path, os.path.basename(test_c_file)) + '.g' + ext
+        ext = 'flag'
+        cbasename = os.path.basename(utils.strip_extension(test_c_file))
+        outtest1 = os.path.join(xrfi_path, cbasename) + '.x.' + ext
```
mkolopanis: there are a lot of lines you call something like: ` '.'.join([os.path.join(xrfi_path, basename), ext])` but this line and the next are different. I mean they both fulfill the same purpose and both work, just feels weird that it switched to a different coding style. Which is understandable since a lot of people have touched this code at some point.

mkolopanis:
also regarding the PEP8 Requirements, it definitely helps with readability. I'm always in favor of adding it, however it may be necessary to have a lot changes like you did with this one.

plaplant:
@mkolopanis agreed. I'd probably make a separate PR once this is merged in, along the lines of what we did with hera_cal (https://github.com/HERA-Team/hera_cal/pull/315). I'd make the whole repo pep8 compliant with `autopep8`, and add a `pycodestyle` line to the travis build.

adampbeardsley:
If hera_cal already automatically checks pep8, I'm happy to add it to qm (when did we let cal get ahead of us on coding practice?!?). But yeah, let's make that a separate PR.

Issue #201, opened 2018-09-11
sequential differencing code produces bad values when pad=True


plaplant:
The `sequential_diff` function in `vis_metrics.py` seems to produce invalid values when `pad=True`. The stated goal of the option is to keep the dimensions of the output the same as the input, but it seems to introduce poorly behaved values as output. The tests on travis show `invalid value encountered in sqrt`, indicating that a negative number was generated prior to the sqrt operation. The root cause of these warnings should be identified, because as it stands we cannot trust the results. If the cause cannot be identified, then the `pad=True` option should be removed.

adampbeardsley:
bump @nkern

plaplant:
Fixed by #225.

PR #225, opened 2018-11-01
fixed runtimewarning in sqrt in vis_metrics.sequential_diff

nkern:
addresses #201

### DIFF
--- a/hera_qm/vis_metrics.py
+++ b/hera_qm/vis_metrics.py
@@ -54,13 +54,10 @@ def check_noise_variance(data):
     return Cij
 
 
-def sequential_diff(data, t_int=None, axis=(0,), pad=True, history=''):
+def sequential_diff(data, t_int=None, axis=(0,), pad=True, run_check=True, history=''):
     """
     Take a sequential (forward) difference of a visibility waterfall
-    as an estimate of the visibility noise. Note: the 1/sqrt(2) correction
-    needed after taking a difference is not applied to the data directly,
-    but is taken into account by the output t_int. If input data is a UVData
-    object, this correction is multiplied into the nsample_array.
+    as an estimate of the visibility noise.
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


hera_qm/vis_metrics.py
```diff
@@ -54,13 +54,10 @@ def check_noise_variance(data):
     return Cij
 
 
-def sequential_diff(data, t_int=None, axis=(0,), pad=True, history=''):
+def sequential_diff(data, t_int=None, axis=(0,), pad=True, run_check=True, history=''):
     """
     Take a sequential (forward) difference of a visibility waterfall
```
plaplant: Can you make this "one-line summary" actually fit on a single line? When the short-form of the docstring is requested, only the first line is returned, so it's helpful for the summary to fit on a single line.

nkern:
I think the negative values were coming from the fact that `true_divide` returns an `empty` array when `where` is False, and the `empty` array had some stochasticness to it, and it sometimes went to like `10^-256` and `-10^-256`, causing the error.

hera_qm/vis_metrics.py
```diff
@@ -138,7 +137,9 @@ def sequential_diff(data, t_int=None, axis=(0,), pad=True, history=''):
             inv_t_int = np.true_divide(1., inv_sum, where=where)
 
             # get noise correction factor
-            corr = np.sqrt(np.true_divide(inv_t_int, t_int, where=where))
+            corr = np.true_divide(inv_t_int, t_int, where=where)
+            corr[np.isclose(corr, 0.0) + (corr < 0) + np.isnan(corr)] = 0.0   # fixes invalid value in sqrt RuntimeWarning
```
plaplant:
```suggestion
            # remove invalid entries from masked values in true_divide
            corr[np.isclose(corr, 0.0) + (corr < 0) + np.isnan(corr)] = 0.0
```

plaplant:
Great! Thanks for doing the leg work. I didn't know that was the behavior of the `where` argument, and that makes sense why we'd get spurious negative values (because as you've found, `empty` makes no guarantees about the contents of the array, unlike `zeros` or `ones`). I've made a suggestion for making the comment a little more precise. If the new version is okay with you, I'm okay with merging this in.

nkern:
Sure, thanks @plaplant !

PR #237, opened 2018-11-15
remove py3 test skips cause hera_cal is py3 compatible now

### LINKED ISSUE #166: Make ant metrics python 3 compatible

### LINKED ISSUE #233: Remove python 3 conditional imports and tests of hera_cal
hera_cal is now python 3 compatible, so we can now remove things like this:

https://github.com/HERA-Team/hera_qm/blob/8fd44f75831926e48687279ea0a179a61deb72d3/hera_qm/tests/test_uvflag.py#L69

adampbeardsley:
fixes #233 
fixes #166


mkolopanis:
Should I still review this? Maybe we should get someone else in the loop

hera_qm/metrics_io.py
```diff
@@ -663,7 +663,9 @@ def _recursively_validate_dict(in_dict):
             in_dict[key] = list(map(tuple, in_dict[key]))
 
         if key in antpol_keys:
-            in_dict[key] = list(map(tuple, np.array(in_dict[key], dtype=antpol_dtype)))
+            # in_dict[key] = list(map(tuple, np.array(in_dict[key], dtype=antpol_dtype)))
```
plaplant: We should remove this line instead of just commenting it out.

hera_qm/metrics_io.py
```diff
@@ -498,7 +498,7 @@ def _parse_list_of_antpols(input_str):
     for entry in entries:
         ant, pol = entry[1:-1].split(",")
         pol = pol[-2]
-        li.append(tuple((int(ant), str(pol))))
+        li.append(tuple((int(ant), qm_utils._str_to_bytes(str(pol)))))
```
plaplant: Why do you need to wrap this in `_str_to_bytes`? Is this list written out to disk at some point? Why isn't the wrapping done as it's being written in that case?

hera_qm/metrics_io.py
```diff
@@ -663,7 +663,9 @@ def _recursively_validate_dict(in_dict):
             in_dict[key] = list(map(tuple, in_dict[key]))
 
         if key in antpol_keys:
-            in_dict[key] = list(map(tuple, np.array(in_dict[key], dtype=antpol_dtype)))
+            # in_dict[key] = list(map(tuple, np.array(in_dict[key], dtype=antpol_dtype)))
+            in_dict[key] = [tuple((ant, qm_utils._bytes_to_str(pol)))
```
plaplant: Same question re: wrapping in `_bytes_to_str`

hera_qm/metrics_io.py
mkolopanis: so this, and the next one, didn't _use_ to have this. But i noticed some weird type casting when I was running the test suite. I wish I had a better answer as to why the types were messed up but this was the only way I found to fix it.

plaplant: Okay, I think we should refactor how these objects are handled then. We really should be doing the type conversion at the I/O interface, not buried in the dictionary validation code or list parsing code. I'm of the opinion that while these objects are in memory, they should be `str` in both python2 and python3, so no funny business should happen in these functions.

mkolopanis: I would definitely prefer that they all remain `str` types too. I'll give this stuff another look.

PR #238, opened 2018-11-15
fix some bookkeeping in collapse_pol, add tests

### LINKED ISSUE #227: UVFlag.collapse_pol results in mixed mode object
For example, I ran:
`uvf.collapse_pol()`
which resulted in a flag mode, but the flag_array is a float (because it did a quadmean across pol). If an average is done across pol, it needs to result in metric mode. If it does an OR it should remain in flag mode.

adampbeardsley:
fixes #227

Whoever reviews - could you double check that I tested the right combinations. I was able to make a test that broke before the patch and now passes. But it's possible I missed other combinations that should be tested.

### DIFF
--- a/hera_qm/tests/test_uvflag.py
+++ b/hera_qm/tests/test_uvflag.py
@@ -614,6 +614,25 @@ def test_collapse_pol():
     uvf2.collapse_pol()
     nt.assert_true(len(uvf2.polarization_array) == 1)
     nt.assert_true(uvf2.polarization_array[0] == ','.join(map(str, uvf.polarization_array)))
+    nt.assert_equal(uvf2.mode, 'metric')
+    nt.assert_true(hasattr(uvf2, 'metric_array'))
+    nt.assert_false(hasattr(uvf2, 'flag_array'))
+
+
+def test_collapse_pol_or():
+    uvf = UVFlag(test_f_file)
+    uvf.to_flag()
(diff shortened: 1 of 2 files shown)


hera_qm/tests/test_uvflag.py
```diff
@@ -614,6 +614,25 @@ def test_collapse_pol():
...
+    uvf = UVFlag(test_f_file)
+    uvf.to_flag()
+    uvf.weights_array = np.ones_like(uvf.weights_array)
+    uvf2 = uvf.copy()
+    uvf2.polarization_array[0] = -4
+    uvf.__add__(uvf2, inplace=True, axis='pol')  # Concatenate to form multi-pol object
+    uvf2 = uvf.copy()
+    uvf2.collapse_pol(method='or')
+    nt.assert_true(len(uvf2.polarization_array) == 1)
+    nt.assert_true(uvf2.polarization_array[0] == ','.join(map(str, uvf.polarization_array)))
```
plaplant: Is this behavior documented? That the polarization array becomes a string when you `or` along it? (I think that's a fine design choice, just making sure we documented that somewhere.)

plaplant: Relatedly, special handling needs to happen for the I/O of this attribute depending on whether it's a string or integers. I'll make an issue.

adampbeardsley: it's documented... now.

PR #245, opened 2018-12-20
fix lst_array in to_waterfall function

adampbeardsley:
Fixes oversight where lst_array was not getting updated in UVFlag object when going to waterfall type.


plaplant:
This PR is failing because of a change in the latest version of h5py that now throws `DeprecationWarnings` when using the `dataset.value` syntax. We'll probably need a new PR to fix that (like https://github.com/RadioAstronomySoftwareGroup/pyuvdata/pull/512), then rebase this branch when that's done.

PR #246, opened 2019-01-03
Ant metrics all pols

### LINKED ISSUE #217: ant_metrics_run doesn't work for files containing all pols
Right now `ant_metrics_run` assumes the files given are single polarization and it goes off and finds the other pols assuming a filename structure. For H2C we need to be able to input files that already contain all pols.
I think we can fix this by making `utils.generate_fullpol_file_list()` smart enough. Alternatively we could just add a keyword to bypass looking for other files.

mkolopanis:
Added comparison in `utils.generate_fullpol_filelist` of the polarizations in the input files to the desired polarizations.

- If the inputs have all the polarizations you want it returns the input files.
- if the inputs have some polarizations (e.g. a file with xx and yy but you want all four pols) it errors 
    - I think this a fringe case an the user should expect only to have a subset of pols and use that as the input to `generate_fullpol_filelist`
- if inputs only have one pol it uses the old regex method to search for remaining polarization files

Currently it uses a `reduce` call to compare the input polarization array, the expected polarization array, and a variable used in the final comparison. This can easily be changed to loop or separate statements if we think it is unclear what is happening.

Closes #217


hera_qm/utils.py
```diff
@@ -332,6 +333,45 @@ def generate_fullpol_file_list(files, pol_list):
...
+            # if a file has more than one polarization but not all expected pols
+            # raise an error that mixed pols are not allowed.
+            if len(input_pols) > 1:
+                raise ValueError("The file: {fname} contains {npol} "
+                                 "polarizations: {pol}."
+                                 "Currently only full lists of all expected "
+                                 "polarization files or lists of "
+                                 "files with single polarizations in the "
+                                 "name of the file (e.g. zen.JD.pol.HH.uv) "
+                                 "are allowed.")
```
adampbeardsley: Should there be a `.format()` at the end of this string?

hera_qm/utils.py
```diff
@@ -332,6 +333,45 @@ def generate_fullpol_file_list(files, pol_list):
...
+
+            else:
+                # if only one polarization then try the old regex method
+                # assumes all files have the same number of polarizations
+                continue
```
adampbeardsley: I think you can do break here to save time.

hera_qm/utils.py
mkolopanis: Why yes, yes there should.

hera_qm/utils.py
mkolopanis: good catch. I probably was thinking I wanted the functionality of break anyway

PR #249, opened 2019-01-07
teach calculate_metric to handle total_quality_array

adampbeardsley:


hera_qm/xrfi.py
```diff
@@ -506,7 +506,7 @@ def flag_apply(uvf, uv, keep_existing=True, force_pol=False, history='',
 # Higher level functions that loop through data to calculate metrics
 #############################################################################
 
-def calculate_metric(uv, algorithm, gains=True, chisq=False, **kwargs):
+def calculate_metric(uv, algorithm, gains=True, chisq=False, tot_chisq=False, **kwargs):
```
jsdillon: Can we combine all three kwargs into one that takes a string?

PR #254, opened 2019-01-31
Firstcal multi pol

### LINKED ISSUE #219: Make sure firstcal metrics can read 4-pol firstcal files

adampbeardsley:
Do not merge until we all agree on how we're handling the plotting code

Closes #219


adampbeardsley:
For the plotting code:
matplotlib should _not_ be a top level requirement.
if we want to keep plotting, matplotlib should be imported in the functions
if matplotlib doesn't exist, just skip tests.
make `hera_qm.plot` module

mkolopanis:
So here is what I have done: 
- Separated the GPR fits to be done per polarization
- Used the pyuvdata `skip` function from pyuvdata.tests to make a `skipIf_no_matplotlib` decorator.
- Moved all matplotlib imports inside of plotting functions and tests.

I am beginning to think that the moving of plots should be a separate PR and merging this current PR in will at least get firstcal metrics ready for RTP with multiple polarization files.

hera_qm/firstcal_metrics.py
```diff
@@ -402,8 +371,8 @@ def __init__(self, calfits_files, use_gp=True):
 
         # Calculate avg delay solution and subtract to get delay_fluctuations
         self.delays = self.delays * 1e9
-        self.delay_avgs = np.median(self.delays, axis=1)
-        self.delay_fluctuations = (self.delays.T - self.delay_avgs).T
+        self.delay_avgs = np.median(self.delays, axis=1, keepdims=True)
+        self.delay_fluctuations = (self.delays - self.delay_avgs)
 
         # use gaussian process model to subtract underlying mean function
         if use_gp is True and self.sklearn_import is True:
```
plaplant: Do we want to expose the Gaussian process parameters/kernel choices to the user? Seems like things have been fine without doing that, but I didn't know if we should add these choices as kwargs.

hera_qm/firstcal_metrics.py
```diff
@@ -349,15 +320,8 @@ def __init__(self, calfits_files, use_gp=True):
         self.UVC = UVCal()
         self.UVC.read_calfits(calfits_files)
 
-        if len(self.UVC.jones_array) > 1:
-            raise ValueError('Sorry, only single pol firstcal solutions are '
```
plaplant: Hooray for removing errors because we've added functionality! 😄

hera_qm/firstcal_metrics.py
```diff
@@ -298,17 +266,20 @@ def plot_zscores(metrics, fname=None, plot_type='full', ax=None, figsize=(10, 6)
...
+    """Object to store and compute FirstCal metric data.
+
     FirstCal_Metrics class for holding firstcal data,
     running metrics, and plotting delay solutions.
     Currently only supports single polarization solutions.
     """
 
     # sklearn import statement
     sklearn_import = sklearn_import
+    jones2pol = {-5: 'XX', -6: 'YY', -7: 'XY', -8: 'YX'}
```
plaplant: Should we import these from pyuvdata instead? Just to make sure that we're being totally self-consistent?

hera_qm/firstcal_metrics.py
mkolopanis: :tada:

hera_qm/firstcal_metrics.py
```diff
@@ -382,17 +347,21 @@ def __init__(self, calfits_files, use_gp=True):
...
             gain_slope = (d_phi / d_nu)
             self.delays = gain_slope / (-2 * np.pi)
             self.gains = fc_gains
 
             # get delay offsets at nu = 0 Hz, and then get rotated antennas
-            self.offsets = fc_phi[:, :, 0] - gain_slope * freqs[0]
-            self.rot_ants = np.unique(list(map(lambda x: self.ants[x], (np.isclose(np.pi, np.abs(self.offsets) % (2 * np.pi), atol=1.0)).T))).tolist()
+            self.offsets = fc_phi[:, :, 0, :] - gain_slope * freqs[0]
+            # find where the offest have a difference of pi from 0
+            rot_offset_bool = np.isclose(np.pi, np.mod(np.abs(self.offsets), 2 * np.pi), atol=1.0).T
```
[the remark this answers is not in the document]
mkolopanis: I think this is tolerance is a (wait for it) _legacy_ carryover. I haven't personally investigated it. But as you bring it up, it does sound like a large tolerance for something like this.

hera_qm/firstcal_metrics.py
```diff
@@ -382,17 +347,21 @@ def __init__(self, calfits_files, use_gp=True):
...
+            rot_offset_bool = np.any(rot_offset_bool, axis=(0, 1))
+            self.rot_ants = np.unique(self.ants[rot_offset_bool])
+            # self.rot_ants = np.unique(list(map(lambda x: self.ants[x], (np.isclose(np.pi, np.abs(self.offsets) % (2 * np.pi), atol=1.0)).T))).tolist()
```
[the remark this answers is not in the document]
mkolopanis: yes, yes it should.

hera_qm/firstcal_metrics.py
mkolopanis: That's a good idea.

hera_qm/firstcal_metrics.py
mkolopanis: sounds like a good project for an interested grad student :+1:

hera_qm/firstcal_metrics.py
```diff
@@ -462,101 +440,85 @@ def run_metrics(self, std_cut=0.5):
...
+            metrics['rot_ants'] = self.rot_ants
+
+            if self.history != '':
+                metrics['history'] = self.history
+
+            full_metrics[pol] = metrics
+
+        self.metrics = full_metrics
+
+    def write_metrics(self, filename=None, filetype='json', overwrite=False):
```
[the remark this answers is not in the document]
mkolopanis: :+1:

hera_qm/firstcal_metrics.py
```diff
@@ -643,7 +607,7 @@ def delay_std(self, return_dict=False):
                 z_scores_d[ant] = z_scores[i]
                 ant_z_scores_d[ant] = ant_z_scores[i]
             for i, t in enumerate(self.times):
-                time_std_d[t] = time_std[i]
+                time_std_d[str(t)] = time_std[i]
```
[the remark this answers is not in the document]
mkolopanis: Yeah the fear was losing JD information to floating point truncation or rounding :thinking: good idea.

hera_qm/firstcal_metrics.py
```diff
@@ -777,10 +742,9 @@ def plot_delays(self, ants=None, plot_type='both', cmap='nipy_spectral', ax=None
         if custom_ax is False:
             return fig
 
-    def plot_zscores(self, fname=None, plot_type='full', ax=None, figsize=(10, 6),
+    def plot_zscores(self, fname=None, plot_type='full', pol=None, ax=None, figsize=(10, 6),
```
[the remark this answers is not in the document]
mkolopanis: I agree with the rename of the class, but was not sure how many other things (possibly up/down stream could break) because of it. That being said, I _think_ the only place that calls it is an jupyter notebook so it probably wouldn't really matter.

I'm not very knowledgeable about why there are these kinds of "namespace pollution" as you call it, but I agree it would be better to write it out in a new PR.

plots/firstcal_metrics_plots.py
```diff
@@ -0,0 +1,354 @@
+# -*- coding: utf-8 -*-
```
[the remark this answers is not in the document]
mkolopanis: I actually thought I removed this file during a revert. I wonder why it is back.

hera_qm/metrics_io.py
```diff
@@ -33,7 +38,7 @@
 known_string_keys = ['history', 'version', 'filedir', 'cut_edges',
                      'fc_filename', 'filename', 'fc_filestem', 'filestem',
                      'pol', 'ant_pol', 'chisq_good_sol', 'good_sol',
-                     'ant_phs_std_good_sol']
+                     'ant_phs_std_good_sol', 'test_string_scalar']
```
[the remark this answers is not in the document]
mkolopanis: Hm it may have been me being a little lazy when trying to implement something. I'll go back and take another look

hera_qm/tests/__init__.py
```diff
@@ -47,12 +49,29 @@ def recursive_compare_dicts(d1, d2):
...
+                print('d2:', key, 'key value type:', type(d1[key]),
+                      'data type:', type(d2[key][0]), d2[key])
+                raise err
+
+
+def skipIf_no_matplotlib(test_func):
+    """Define a decorator to skip tests that require matplotlib."""
+    reason = 'matplotlib is not installed, skipping tests that require it.'
+    try:
+        import healpy
```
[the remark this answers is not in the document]
mkolopanis: the dangers of copy/pasta

hera_qm/firstcal_metrics.py
```diff
@@ -382,17 +346,23 @@ def __init__(self, calfits_files, use_gp=True):
         if self.UVC.cal_type == 'gain':
             # get delays
             freqs = self.UVC.freq_array.squeeze()
-            fc_gains = np.moveaxis(self.UVC.gain_array, 2, 3)[:, 0, :, :, 0]
-            fc_phi = np.unwrap(np.angle(fc_gains))
+            # Readers/writers beware!
+            # The time an frequency axes are being swapped here
```
plaplant: It seems like this is no longer true (which is amazing--thanks!)

hera_qm/firstcal_metrics.py
```diff
@@ -420,20 +390,30 @@ def __init__(self, calfits_files, use_gp=True):
...
-                ystd = np.sqrt(astats.biweight_midvariance(y))
-                y /= ystd
-                # fit GP and remove from delay fluctuations
-                GP = gp.GaussianProcessRegressor(kernel=kernel, n_restarts_optimizer=0)
-                GP.fit(x, y)
-                ymodel = GP.predict(x).ravel() * ystd
-                self.delay_fluctuations[i] -= ymodel
-                self.delay_smooths.append(ymodel)
+                ystd = np.sqrt(astats.biweight_midvariance(y, axis=0))
+                # if this is a simulation it is possible all the delays are 1
```
plaplant: Thanks for making this clearer 👍

hera_qm/firstcal_metrics.py
```diff
@@ -297,9 +265,10 @@ def plot_zscores(metrics, fname=None, plot_type='full', ax=None, figsize=(10, 6)
         return fig
 
 
-class FirstCal_Metrics(object):
-    """
-    FirstCal_Metrics class for holding firstcal data,
+class FirstCalMetrics(object):
```
plaplant: Neato, thanks for making this change

jsdillon:
nice work! Wanna add this to the IDR2.2 opm?

PR #256, opened 2019-02-27
make uvflag inherit object

mwilensky768:
`UVFlag` now properly inherits `object` as in 

`class UVFlag(object):`

Seems to pass integration checks on push.


mkolopanis:
ah interesting, this makes little difference in python 3 but definitely something we should have for python 2 users

adampbeardsley:
Agreed. It's ready to go, except I asked @mwilensky768 to make a quick test to prove it does what it's supposed to.

PR #258, opened 2019-03-05
xrfi for IDR2 2

adampbeardsley:
This is the PR to implement the new xrfi strategy for IDR2.2. We've talked about this strategy and some of the changes on QM and Analysis telecons. There are a lot of changes, so I'll go through and do a "review" myself to point out some of the bits reviewers will want to notice.


LICENSE
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2018 Hydrogen Epoch of Reionization Array (HERA)
+Copyright (c) 2019 Hydrogen Epoch of Reionization Array (HERA)
```
adampbeardsley: I updated the copyright dates while I was waiting for tests to run. Sorry reviewers.

hera_qm/tests/test_utils.py
```diff
@@ -121,8 +121,8 @@ def test_get_metrics_ArgumentParser_omnical_metrics():
     nt.assert_equal(args.extension, 'foo')
 
 
-def test_get_metrics_ArgumentParser_xrfi_run():
-    a = utils.get_metrics_ArgumentParser('xrfi_run')
+def test_get_metrics_ArgumentParser_xrfi_h1c_run():
+    a = utils.get_metrics_ArgumentParser('xrfi_h1c_run')
```
adampbeardsley: The old `xrfi_run` is now labeled `xrfi_h1c_run`. The general idea is that current strategies get the privilege of simple names, and legacy ones get labels indicating which analysis they were used for.

hera_qm/tests/test_xrfi.py
```diff
@@ -120,7 +131,7 @@ def test_medminfilt(self):
         # test cases where filters are larger than data dimensions
         Kt = self.size + 1
         Kf = self.size + 1
-        d_filt = uvtest.checkWarnings(xrfi.medminfilt, [data, Kt, Kf], nwarnings=2,
+        d_filt = uvtest.checkWarnings(xrfi.medminfilt, [data], {'Kt': Kt, 'Kf': Kf}, nwarnings=2,
```
adampbeardsley: There were a bunch of places we were using positional arguments for things that should be keywords.

hera_qm/utils.py
```diff
@@ -635,9 +687,42 @@ def or_collapse(a, weights=None, axis=None, returned=False):
         return o
 
 
+def and_collapse(a, weights=None, axis=None, returned=False):
```
adampbeardsley: This function allows us to convert a flag array to waterfall by flagging time/freq/pol pixels that are completely flagged.

hera_qm/utils.py
```diff
@@ -635,9 +687,42 @@ def or_collapse(a, weights=None, axis=None, returned=False):
...
+    o = np.all(a, axis=axis)
+    if (weights is not None) and not np.all(weights == weights.reshape(-1)[0]):
+        warnings.warn('Currently weights are not handled when AND-ing boolean arrays.')
+    if returned:
+        return o, np.ones_like(o, dtype=np.float)
+    else:
+        return o
+
+
+def and_rows_cols(waterfall):
```
adampbeardsley: This is Nick's function that enables his convert_wgts option when converting from flag to metric UVFlag object

hera_qm/uvflag.py
```diff
@@ -473,7 +511,7 @@ def to_waterfall(self, method='quadmean', keep_pol=True):
             self.weights_array = w
             self.time_array, ri = np.unique(self.time_array, return_index=True)
             self.lst_array = self.lst_array[ri]
-        if (method == 'or') and (self.mode == 'flag'):
+        if ((method == 'or') or (method == 'and')) and (self.mode == 'flag'):
```
adampbeardsley: Here and the next two bits were bugs

hera_qm/version.py
```diff
@@ -1,36 +1,55 @@
...
+# Copyright (c) 2019 the HERA Project
 # Licensed under the MIT License
 
 import os
 import subprocess
 import json
+import six
+
+
+hera_qm_dir = os.path.dirname(os.path.realpath(__file__))
```
adampbeardsley: Most of the changes in this file are getting the str/bytes stuff to work in py3. I don't understand it... mostly taken from pyuvdata.

hera_qm/xrfi.py
```diff
@@ -70,14 +70,32 @@ def flag_xants(uv, xants, inplace=True):
         return uvo
 
 
+def resolve_xrfi_path(xrfi_path, file):
```
adampbeardsley: convenience function

hera_qm/xrfi.py
```diff
@@ -70,14 +70,32 @@ def flag_xants(uv, xants, inplace=True):
...
+        dirname = os.path.dirname(os.path.abspath(file))
+    return dirname
+
+
 #############################################################################
 # Functions for preprocessing data prior to RFI flagging
 #############################################################################
 
-def medmin(d):
+def medmin(d, flags=None):
```
adampbeardsley: To keep the interface to these functions uniform, I added a `flags` keyword to all of them, even if it's not used in most of them.

hera_qm/xrfi.py
```diff
@@ -222,9 +251,52 @@ def detrend_medfilt(d, Kt=8, Kf=8):
     return f[Kt:-Kt, Kf:-Kf]
 
 
+def detrend_meanfilt(d, flags=None, Kt=8, Kf=8):
```
adampbeardsley: This is our first filtering function that can actually use the flags to mask known outliers. I want to build on this to make a median filter that can do the same, but I couldn't find a suitable out-of-box solution so I started with this.

hera_qm/xrfi.py
```diff
@@ -410,13 +482,13 @@ def flag(uvf_m, nsig_p=6., nsig_f=None, nsig_t=None, avg_method='quadmean'):
 
     # Pixel flagging
     if nsig_p is not None:
-        uvf_f.flag_array[uvf_m.metric_array > nsig_p] = True
+        uvf_f.flag_array[uvf_m.metric_array >= nsig_p] = True
```
adampbeardsley: The "or equal" allows us to use 1 as a threshold when flagging on fraction of antenna/baselines that flagged a specific pixel.

hera_qm/xrfi.py
```diff
@@ -607,11 +685,215 @@ def xrfi_h1c_pipe(uv, Kt=8, Kf=8, sig_init=6., sig_adj=2., px_threshold=0.2,
...
+            to a flag. Default is 2.0.
+    Returns:
+        uvf_m (UVFlag): UVFlag object with metric after collapsing to waterfall and to
+            single pol. Weights array is set to ones.
+        uvf_fws (UVFlag): UVFlag object with flags after watershed.
+    """
+    flag_xants(uv, xants)
+    uvf_m = calculate_metric(uv, alg, Kt=Kt, Kf=Kf, cal_mode=cal_mode)
+    uvf_m.to_waterfall(keep_pol=False)
+    uvf_m.weights_array = uvf_m.weights_array.astype(np.bool).astype(np.float)
```
adampbeardsley: This resets the weights to 1 (with data) or 0 (no data) to equally combine with the other metrics.

hera_qm/xrfi.py
```diff
@@ -607,11 +685,215 @@ def xrfi_h1c_pipe(uv, Kt=8, Kf=8, sig_init=6., sig_adj=2., px_threshold=0.2,
...
+    alg_func = algorithm_dict[alg]
+    uvf_m.metric_array[:, :, 0] = alg_func(uvf_m.metric_array[:, :, 0],
+                                           flags=~uvf_m.weights_array[:, :, 0].astype(np.bool),
+                                           Kt=Kt, Kf=Kf)
```
adampbeardsley: This is the renormalization step where we pass the z-scores through the filter again to get a zero-centered, width-of-one distribution.

hera_qm/xrfi.py
```diff
@@ -607,11 +685,215 @@ def xrfi_h1c_pipe(uv, Kt=8, Kf=8, sig_init=6., sig_adj=2., px_threshold=0.2,
...
+    uvf_f = flag(uvf_m, nsig_p=sig_init)
+    uvf_fws = watershed_flag(uvf_m, uvf_f, nsig_p=sig_adj, inplace=False)
```
adampbeardsley: Flag and watershed on each data product individually. That is, on each complete file (e.g. calibration gains), not on individual antennas/baselines.
We don't broadcast until the very end.

hera_qm/xrfi.py
```diff
@@ -607,11 +685,215 @@ def xrfi_h1c_pipe(uv, Kt=8, Kf=8, sig_init=6., sig_adj=2., px_threshold=0.2,
...
+    uvf_metrics.metric_array[:, :, 0] = alg_func(uvf_metrics.metric_array[:, :, 0],
+                                                 flags=~uvf_metrics.weights_array[:, :, 0].astype(np.bool),
+                                                 Kt=kt_size, Kf=kf_size)
+
+    # Flag on combined metrics
+    uvf_f = flag(uvf_metrics, nsig_p=sig_init)
+    uvf_fws = watershed_flag(uvf_metrics, uvf_f, nsig_p=sig_adj, inplace=False)
+    # OR everything together for initial flags
+    uvf_apriori.to_waterfall(method='and', keep_pol=False)
+    uvf_init = uvf_fws | uvf_ogf | uvf_oxf | uvf_agf | uvf_axf | uvf_vf | uvf_apriori
```
adampbeardsley: This combination of flags becomes the initial mask for the second round, which includes the raw visibility data.

hera_qm/xrfi.py
```diff
@@ -607,11 +685,215 @@ def xrfi_h1c_pipe(uv, Kt=8, Kf=8, sig_init=6., sig_adj=2., px_threshold=0.2,
...
+    # Flag on combined metrics
+    uvf_f2 = flag(uvf_metrics2, nsig_p=sig_init)
+    uvf_fws2 = watershed_flag(uvf_metrics2, uvf_f2, nsig_p=sig_adj, inplace=False)
+    uvf_combined2 = (uvf_fws2 | uvf_ogf2 | uvf_oxf2 | uvf_agf2 | uvf_axf2
+                     | uvf_vf2 | uvf_df2 | uvf_init)
+
+    # Threshold
+    uvf_temp = uvf_combined2.copy()
+    uvf_temp.to_metric(convert_wgts=True)
+    uvf_final = flag(uvf_temp, nsig_p=1.0, nsig_f=freq_threshold, nsig_t=time_threshold)
```
adampbeardsley: This is the only place we threshold/broadcast - at the very end

hera_qm/utils.py
```diff
@@ -635,9 +687,42 @@ def or_collapse(a, weights=None, axis=None, returned=False):
         return o
 
 
+def and_collapse(a, weights=None, axis=None, returned=False):
+    ''' Function to collapse axes using AND operation
+    Args:
+        a - boolean array to process
+        weights - NOT USED, but kept for symmetry with other averaging functions
+        axis - axis or axes over which to OR
```
jsdillon: Don't you mean AND?

hera_qm/utils.py
jsdillon: I question the need for this function. On one hand, it's symmetric with with other steps. However the guts of it are basically just one line, `o = np.all(a, axis=axis)`. In the interests of code maintainability and avoiding bloat, I'd replace it in the code with the numpy equivalent.

hera_qm/utils.py
```diff
@@ -635,9 +687,42 @@ def or_collapse(a, weights=None, axis=None, returned=False):
...
+    else:
+        return o
+
+
+def and_rows_cols(waterfall):
+    """ For a 2D flag waterfall, flag pixels only if fully flagged along
+    time and/or frequency
+    Args:
+        waterfall - 2D boolean array of shape (Ntimes, Nfreqs)
+    """
```
jsdillon: This needs to explain the returned value.

hera_qm/utils.py
```diff
@@ -635,9 +687,42 @@ def or_collapse(a, weights=None, axis=None, returned=False):
...
+    wf = np.zeros_like(waterfall, dtype=np.bool)
+    Ntimes, Nfreqs = waterfall.shape
+    wf[:, (np.sum(waterfall, axis=0) / Ntimes) == 1] = True
+    wf[(np.sum(waterfall, axis=1) / Nfreqs) == 1] = True
+    return wf
+
+
 # Dictionary to map different methods for averaging data.
 averaging_dict = {'mean': mean, 'absmean': absmean, 'quadmean': quadmean,
```
jsdillon: Structurally, I find this a bit weird. You've gone out of your way to make these all have the same input format. I think you'd save a lot of complexity if you just combined them all into one function with a method kwarg.

hera_qm/uvflag.py
```diff
@@ -429,6 +429,44 @@ def copy(self):
...
+        else:
+            this = self.copy()
+        method = method.lower()
+        avg_f = qm_utils.averaging_dict[method]
+        darray = np.expand_dims(this.metric_array, 0)
+        warray = np.expand_dims(this.weights_array, 0)
+        for other in others:
+            if this.metric_array.shape != other.metric_array.shape:
+                raise ValueError('UVFlag metric arrays do not match.')
+            darray = np.vstack([darray, np.expand_dims(other.metric_array, 0)])
```
jsdillon: It doesn't matter a ton, but looping over vstack is slower than vstacking everything at once.

hera_qm/uvflag.py
jsdillon: I'm having trouble interpreting these cases. Is the idea that if you use the other averaging methods when you're in flag mode, you'll go into metric mode? Is that the most sensible behavior? Perhaps explain that in comments.

hera_qm/uvflag.py
```diff
@@ -630,20 +668,36 @@ def to_flag(self):
...
 
-    def to_metric(self):
-        '''Convert to metric mode. NOT SMART. Simply removes flag_array and initializes
-        metric_array with zeros.
+    def to_metric(self, convert_wgts=False):
+        '''Convert to metric mode. NOT SMART. Simply recasts flag_array as float
+        and uses this as the metric array.
+
+        Args:
+            convert_wgts : bool, if True convert self.weights_array to ones
```
jsdillon: Why would one want to do this? Perhaps explain in the docstring.

hera_qm/uvflag.py
```diff
@@ -630,20 +668,36 @@ def to_flag(self):
...
         elif self.mode == 'flag':
-            self.metric_array = np.zeros_like(self.flag_array, dtype=np.float)
+            self.metric_array = self.flag_array.astype(np.float)
             self.mode = 'metric'
+            if convert_wgts:
+                self.weights_array = np.ones_like(self.weights_array)
+                if self.type == 'waterfall':
+                    for i, pol in enumerate(self.polarization_array):
+                        self.weights_array[:, :, i] *= ~qm_utils.and_rows_cols(self.flag_array[:, :, i])
+                elif self.type == 'baseline':
```
jsdillon: Do we not care about self.type == 'antenna'?

hera_qm/version.py
```diff
@@ -61,6 +80,7 @@ def construct_version_info():
                             + '.  Git hash: ' + git_hash
                             + '.  Git branch: ' + git_branch
                             + '.  Git description: ' + git_description + '.')
+hera_qm_version_str = str(hera_qm_version_str)
```
jsdillon: You might be interested in this nicely formatted history string function I produced for hera_cal: https://github.com/HERA-Team/hera_cal/blob/df1009b80773cb8c304accc92f5e5c8c0d5b482b/hera_cal/version.py#L84

Here's an example: 
```
------------
This file was produced by the function post_redcal_abscal_run() in abscal.py using:
    git_branch: master
    git_description: v2.0-603-gdf1009b
    git_hash: df1009b80773cb8c304accc92f5e5c8c0d5b482b
    git_origin: https://github.com/HERA-Team/hera_cal.git
    version: 2.0
------------
```

hera_qm/xrfi.py
```diff
@@ -70,14 +70,32 @@ def flag_xants(uv, xants, inplace=True):
...
+def resolve_xrfi_path(xrfi_path, file):
+    """ Determine xrfi_path based on given directory or default to dirname of given file.
+    Args:
+        xrfi_path (str): Directory to write xrfi outputs.
+        file (str): Filename to determine backup directory if xrfi_path == ''
+    Returns:
+        dirname (str): If xrfi_path != '', returns xrfi_path. Otherwise returns
+            directory of file.
+    """
+    if xrfi_path != '':
```
jsdillon: Would it make sense to make this more general? Perhaps check whether the directory exists?

hera_qm/xrfi.py
jsdillon: I think it's considered bad style to call a variable `file`, since it's already a build in function.

hera_qm/xrfi.py
```diff
@@ -161,33 +183,40 @@ def detrend_deriv(d, dt=True, df=True):
...
+            NOT USED in this function.
         Kt (int): size in time to apply medminfilter over
         Kf (int): size in frequency to apply medminfilter over
     Returns:
         float array: float array of outlier significance metric
     """
     if d.ndim != 2:
         raise ValueError('Input to detrend_medminfilt must be 2D array.')
-    d_sm = medminfilt(np.abs(d), 2 * Kt + 1, 2 * Kf + 1)
+    d_sm = medminfilt(np.abs(d), Kt=2 * Kt + 1, Kf=2 * Kf + 1)
```
jsdillon: For readability, I generally prefer to put arguments that involve spaces in parens, e.g. `Kt=(2 * Kt + 1)`

hera_qm/xrfi.py
```diff
@@ -222,9 +251,52 @@ def detrend_medfilt(d, Kt=8, Kf=8):
...
+        flags (array, optional): 2D flag array to be interpretted as mask for d.
+        Kt (int, optional): box size in time (first) dimension to apply medfilt
+            over. Default is 8 pixels.
+        Kf (int, optional): box size in frequency (second) dimension to apply medfilt
+            over. Default is 8 pixels.
+    Returns:
+        f: array of outlier significance metric. Same type and size as d.
+    """
+    # Delay import so astropy is not required for any use of hera_qm
+    from astropy.convolution import convolve
```
jsdillon: You may want to mention why you're using astropy rather than scipy's convolve. Maybe just link here: http://docs.astropy.org/en/stable/convolution/

hera_qm/xrfi.py
```diff
@@ -222,9 +251,52 @@ def detrend_medfilt(d, Kt=8, Kf=8):
...
+
+    if d.ndim != 2:
```
jsdillon: These three if statements get run a lot. I'd recommend condensing them into a single `_check_convolve_dims(d, Kt, Kf)`.

hera_qm/xrfi.py
```diff
@@ -222,9 +251,52 @@ def detrend_medfilt(d, Kt=8, Kf=8):
...
+    if Kt > d.shape[0]:
+        warnings.warn("Kt value {0:d} is larger than the data of dimension {1:d}; "
+                      "using the size of the data for the kernel size".format(Kt, d.shape[0]))
+        Kt = d.shape[0]
+    if Kf > d.shape[1]:
+        warnings.warn("Kf value {0:d} is larger than the data of dimension {1:d}; "
+                      "using the size of the data for the kernel size".format(Kf, d.shape[1]))
+        Kf = d.shape[1]
+    kernel = np.ones((2 * Kt + 1, 2 * Kf + 1))
+    d = np.concatenate([d[Kt - 1::-1], d, d[:-Kt - 1:-1]], axis=0)
```
jsdillon: I believe what you're trying to do here is the same as convolve's `boundary='wrap'`. 

However, are you sure you want to wrap and not mirror across the edges of the data?

hera_qm/xrfi.py
```diff
@@ -222,9 +251,52 @@ def detrend_medfilt(d, Kt=8, Kf=8):
...
+    d = np.concatenate([d[:, Kf - 1::-1], d, d[:, :-Kf - 1:-1]], axis=1)
+    if flags is not None:
+        flags = np.concatenate([flags[Kt - 1::-1], flags, flags[:-Kt - 1:-1]], axis=0)
```
jsdillon: Ditto.

hera_qm/xrfi.py
jsdillon: I'd add a comment that you're trying to effectively do a `mirror` extend, like in scipy's convolve, but that astropy doesn't support `mirror`.

hera_qm/xrfi.py
```diff
@@ -222,9 +251,52 @@ def detrend_medfilt(d, Kt=8, Kf=8):
...
+        flags = np.concatenate([flags[:, Kf - 1::-1], flags, flags[:, :-Kf - 1:-1]], axis=1)
+    d_sm = convolve(d, kernel, mask=flags, boundary='extend')
+    d_rs = d - d_sm
+    d_sq = np.abs(d_rs)**2
+    # puts median on same scale as average
```
jsdillon: median?

hera_qm/xrfi.py
```diff
@@ -222,9 +251,52 @@ def detrend_medfilt(d, Kt=8, Kf=8):
...
+    sig = np.sqrt(convolve(d_sq, kernel, mask=flags))
+    # don't divide by zero, instead turn those entries into +inf
+    f = np.true_divide(d_rs, sig, where=(np.abs(sig) > 1e-8))
```
jsdillon: Where does this magic 1e-8 number come from? What if the whole data set is small? Shouldn't it be about whether it's small relative to something?

hera_qm/xrfi.py
```diff
@@ -222,9 +251,52 @@ def detrend_medfilt(d, Kt=8, Kf=8):
...
+    f = np.where(np.abs(sig) > 1e-8, f, np.inf)
+    return f[Kt:-Kt, Kf:-Kf]
+
+
 # Update algorithm_dict whenever new metric algorithm is created.
 algorithm_dict = {'medmin': medmin, 'medminfilt': medminfilt, 'detrend_deriv': detrend_deriv,
```
jsdillon: Again, it feels a little weird to have a dictionary rather than a giant switch. Your call though.

jsdillon: I guess it feels a little less weird here than in the other case, because these are larger functions that are more likely to be used independently and deserve substantial docstrings.

hera_qm/xrfi.py
```diff
@@ -410,13 +482,13 @@ def flag(uvf_m, nsig_p=6., nsig_f=None, nsig_t=None, avg_method='quadmean'):
 
     # Pixel flagging
     if nsig_p is not None:
-        uvf_f.flag_array[uvf_m.metric_array > nsig_p] = True
+        uvf_f.flag_array[uvf_m.metric_array >= nsig_p] = True
 
     if uvf_m.type == 'baseline':
         if nsig_f is not None:
             # Channel flagging
             d = avg_f(uvf_m.metric_array, axis=(0, 1, 3), weights=uvf_m.weights_array)
```
jsdillon: `d` is a confusing variable name. I want it to be data, not an averaged waterfall over some axes.

hera_qm/xrfi.py
```diff
@@ -607,11 +685,215 @@ def xrfi_h1c_pipe(uv, Kt=8, Kf=8, sig_init=6., sig_adj=2., px_threshold=0.2,
     else:
         return uvf_f, uvf_wf
 
+
+def xrfi_pipe(uv, alg='detrend_medfilt', Kt=8, Kf=8, xants=[], cal_mode='gain',
```
jsdillon: Pipe is a bit of strange name for this, since it's still fairly low level in terms of the entire analysis pipeline. Maybe something like `single_object_xrfi`?

hera_qm/xrfi.py
jsdillon: I'd add this as a comment in the code.

hera_qm/xrfi.py
jsdillon: I'd add this as a comment in the code.

hera_qm/xrfi.py
jsdillon: I'd add this as a comment in the code.

hera_qm/xrfi.py
```diff
@@ -607,11 +685,215 @@ def xrfi_h1c_pipe(uv, Kt=8, Kf=8, sig_init=6., sig_adj=2., px_threshold=0.2,
...
+            single pol. Weights array is set to ones.
+        uvf_fws (UVFlag): UVFlag object with flags after watershed.
+    """
+    flag_xants(uv, xants)
+    uvf_m = calculate_metric(uv, alg, Kt=Kt, Kf=Kf, cal_mode=cal_mode)
+    uvf_m.to_waterfall(keep_pol=False)
+    uvf_m.weights_array = uvf_m.weights_array.astype(np.bool).astype(np.float)
+    alg_func = algorithm_dict[alg]
+    uvf_m.metric_array[:, :, 0] = alg_func(uvf_m.metric_array[:, :, 0],
+                                           flags=~uvf_m.weights_array[:, :, 0].astype(np.bool),
```
jsdillon: For clarity, I'd wrap this as `~(uvf_m.weights_array[:, :, 0].astype(np.bool))` to show order of operations.

hera_qm/xrfi.py
```diff
@@ -607,11 +685,215 @@ def xrfi_h1c_pipe(uv, Kt=8, Kf=8, sig_init=6., sig_adj=2., px_threshold=0.2,
...
+            "foo.flagged_abs.calfits".
+    Returns:
+        None
+    """
+    history = 'Flagging command: "' + history + '", Using ' + hera_qm_version_str
+    dirname = resolve_xrfi_path(xrfi_path, data_file)
+    xants = process_ex_ants(ex_ants=ex_ants, metrics_file=metrics_file)
+
+    # Initial run on cal data products
+    alg = 'detrend_medfilt'
```
jsdillon: I don't see the need to make this its own variable unless you want it to be a kwarg. I think it makes the calls to xrfi_pipe less readable.

hera_qm/xrfi.py
```diff
@@ -607,11 +685,215 @@ def xrfi_h1c_pipe(uv, Kt=8, Kf=8, sig_init=6., sig_adj=2., px_threshold=0.2,
...
+
+    # Second round -- use init flags to mask and recalculate everything
+    # Read in data file
+    uv_d = UVData()
+    uv_d.read(data_file)
+    for uv in [uvc_o, uvc_a, uv_v, uv_d]:
+        flag_apply(uvf_init, uv, keep_existing=True, force_pol=True)
+
+    # Do next round of metrics
+    alg = 'detrend_meanfilt'  # Change to meanfilt because it can mask flagged pixels
```
jsdillon: Same comment as before about `alg`

Also, given the amount of repeated code in this function, would it make sense to make a helper function that does omni/abs/omni model flags all together?

hera_qm/xrfi.py
```diff
@@ -607,11 +685,215 @@ def xrfi_h1c_pipe(uv, Kt=8, Kf=8, sig_init=6., sig_adj=2., px_threshold=0.2,
...
+    uvf_final.write(outpath, clobber=True)
+
+    # Save calfits with new flags
+    flag_apply(uvf_final, uvc_a, force_pol=True, history=history)
+    basename = qm_utils.strip_extension(os.path.basename(acalfits_file))
+    basename = qm_utils.strip_extension(basename)  # Also get rid of .abs
+    outfile = '.'.join([basename, cal_ext, 'calfits'])
+    outpath = os.path.join(dirname, outfile)
+    uvc_a.write_calfits(outpath, clobber=True)
+
```
jsdillon: Just wanted to say, even though this function is still quite long, you've done some really great work getting it to the point where it's really truly modular. What an improvement!

hera_qm/tests/test_xrfi.py
```diff
@@ -217,9 +228,64 @@ def test_detrend_medfilt(self):
...
+        for i in range(data.shape[1]):
+            data[:, i] = i * np.ones_like(data[:, i])
+        # run detrend medfilt
+        Kt = 8
+        Kf = 8
+        dm = xrfi.detrend_meanfilt(data, Kt=Kt, Kf=Kf)
+
+        # read in "answer" array
+        # this is output that corresponds to self.size==100, Kt==101, Kf==101
+        ans_fn = os.path.join(DATA_PATH, 'test_detrend_meanfilt_ans.txt')
```
jsdillon: Why do you have to do this? Can't we concoct a test that we know the answer analytically?

hera_qm/uvflag.py
adampbeardsley: Yeah. But when I just rewrote it, I had to keep the loop to check for mismatched shape, then create another variable to keep the expanded dims (or include a long list comprehension). Overall it just got messy, and ultimately this step is very subdominant in terms of time.

hera_qm/xrfi.py
adampbeardsley: To quote @plaplant:
> “we’re doing this song and dance to avoid floating point errors, and this _probably_ won’t lead to an error, but being more rigorous is likely wasted effort”
To add a little to that, it's more complicated than comparing to data because we're trying to avoid overflows due to machine precision. So really we need a comparison of the data exponents against the sigma exponents and threshold against the precision of the data type. I'll code this up a little differently and add a comment.

hera_qm/xrfi.py
adampbeardsley: Yeah, that was the thinking here. I'd like to keep these as-is, but I'll switch the averaging functions.

hera_qm/xrfi.py
adampbeardsley: It's not a pipeline. It's a single pipe. You can string these together if you'd wish to make something like a pipeline. In this instance it does xrfi, but in previous runs (IDR2.1) it did not. I still like the concept of these functions being "pipes" that you can put together.

hera_qm/xrfi.py
adampbeardsley: I thought about that, but the fact that we have one extra (data), and it's only repeated twice, I didn't think it was worth it.

hera_qm/tests/test_xrfi.py
adampbeardsley: In theory. It got to be a pain because of the mirroring. I opted to take the same route that Paul did on other tests and just do it once and the test makes sure you don't break things when updating stuff.

hera_qm/version.py
adampbeardsley: This is a good idea, but I'm going to punt on it for this PR. I only edited the version file because things started failing on travis. Otherwise this is pretty unrelated. I've made an issue so we don't lose track of it.

hera_qm/uvflag.py
adampbeardsley: added

hera_qm/xrfi.py
adampbeardsley: added that functionality.

hera_qm/xrfi.py
adampbeardsley: fixed

hera_qm/xrfi.py
adampbeardsley: fixed

hera_qm/xrfi.py
adampbeardsley: Added to comment

hera_qm/xrfi.py
adampbeardsley: done

hera_qm/xrfi.py
adampbeardsley: done

hera_qm/xrfi.py
adampbeardsley: too much copy-paste. good catch.

hera_qm/xrfi.py
adampbeardsley: added a robust_divide function that uses precision of the actual data.

hera_qm/xrfi.py
adampbeardsley: Fixed. I made a master function which just has a case statement that calls the other functions. I found it useful to still have the other functions because it makes the testing cleaner. And in principle one can use those functions individually.

hera_qm/xrfi.py
adampbeardsley: It's just a throw-away. I don't want to overthink what to call a temporary variable that has no use outside this if statement.

hera_qm/xrfi.py
adampbeardsley: done

hera_qm/xrfi.py
adampbeardsley: Hmm. I actually see the variable as useful in case we decide to change it, there's only one point to change. Nevertheless I took it out of the code for now. Maybe @plaplant can be the deciding vote when he reviews.

hera_qm/xrfi.py
adampbeardsley: Thanks!

jsdillon:
Thanks @adampbeardsley. I'm satisfied.

hera_qm/tests/test_xrfi.py
```diff
@@ -621,17 +686,73 @@ def test_calculate_metric_errors(self):
...
+        nt.assert_equal(uvf_m.type, 'waterfall')
+        nt.assert_equal(len(uvf_m.polarization_array), 1)
+        nt.assert_equal(uvf_m.weights_array.max(), 1.)
 
 
 class TestWrappers():
 
+    def test_xrfi_run(self):
+        # Run in nicest way possible
+        # The warnings are because we use UVFlag.to_waterfall() on the total chisquareds
```
plaplant: Thanks for the helpful comments.

hera_qm/xrfi.py
```diff
@@ -70,14 +70,71 @@ def flag_xants(uv, xants, inplace=True):
...
+                      "using the size of the data for the kernel size".format(Kt, d.shape[0]))
+        Kt = d.shape[0]
+    if Kf > d.shape[1]:
+        warnings.warn("Kf value {0:d} is larger than the data of dimension {1:d}; "
+                      "using the size of the data for the kernel size".format(Kf, d.shape[1]))
+        Kf = d.shape[1]
+    return Kt, Kf
+
+
+def robust_divide(a, b):
```
plaplant: Nice auxiliary function, with a helpful docstring!

hera_qm/uvflag.py
```diff
@@ -454,8 +490,8 @@ def to_waterfall(self, method='quadmean', keep_pol=True):
             darr = self.metric_array
 
         if self.type == 'antenna':
-            d, w = avg_f(darr, axis=(0, 1), weights=self.weights_array,
-                         returned=True)
+            d, w = qm_utils.collapse(darr, method, axis=(0, 1), weights=self.weights_array,
```
plaplant: I'm generally not a huge fan of single-letter variable names. (It's hard to search for them and sometimes what they represent isn't clear.) It seems that you're pretty consistent about `d` being "data" and `w` being "weights", so I'm inclined to let it be (and since it comes up a lot, it'd be difficult/error prone to replace all instances). I think it's just good coding practice to have in the future.

hera_qm/xrfi.py
```diff
@@ -489,6 +563,7 @@ def flag_apply(uvf, uv, keep_existing=True, force_pol=False, history='',
         if f.mode != 'flag':
             raise ValueError('UVFlag objects must be in mode "flag" to apply to data.')
         if f.type == 'waterfall':
+            f = f.copy()  # don't change the input object
```
plaplant: Does this prevent the input object from being changed? I've usually seen a `deepcopy` used to prevent accidental modification, but if `copy()` does the same thing, this is fine.

hera_qm/xrfi.py
```diff
@@ -607,11 +687,220 @@ def xrfi_h1c_pipe(uv, Kt=8, Kf=8, sig_init=6., sig_adj=2., px_threshold=0.2,
...
+    uvf_init.write(outpath, clobber=True)
+
+    # Second round -- use init flags to mask and recalculate everything
+    # Read in data file
+    uv_d = UVData()
+    uv_d.read(data_file)
+    for uv in [uvc_o, uvc_a, uv_v, uv_d]:
+        flag_apply(uvf_init, uv, keep_existing=True, force_pol=True)
+
+    # Do next round of metrics
```
plaplant: Nice descriptive comments.

hera_qm/uvflag.py
adampbeardsley: I'll make an issue to address this throughout the repo.

hera_qm/xrfi.py
adampbeardsley: It's `UVFlag` magic! The `copy` method actually just does a deep copy underneath.

PR #266, opened 2019-03-15
Boolean Io

mkolopanis:
I added boolean as a known type in `metrics_io` no longer casts booleans as strings.


bhazelton:
I can confirm that this fixes the hera_mc test failures.

PR #275, opened 2019-05-04
Fix flag handling in firstcal metrics

jsdillon:
Due to solar flagging, firstcal has some (but not all!) integrations where all antennas are flagged.  This is not something that firstcal_metrics was designed for. This lead to a number of wrong calculations (medians giving 0s, etc.) and errors in the pipeline running right now on files that straddle sunrise and sunset.

This PR fixes that.


hera_qm/firstcal_metrics.py
```diff
@@ -384,30 +388,23 @@ def __init__(self, calfits_files, use_gp=True):
...
-                    y /= ystd
-                    GP = gp.GaussianProcessRegressor(kernel=kernel, n_restarts_optimizer=0)
-                    ymodel_stack = np.zeros((self.Ntimes, self.Npols))
-                    for pol_cnt in range(self.Npols):
-                        # fit GP and remove from delay fluctuations
-                        # but only one polarization at a time
+                ystd = np.sqrt([astats.biweight_midvariance(y[~delay_flags[i, :, ip], ip]) for ip in range(self.Npols)])
+                y /= ystd
+                GP = gp.GaussianProcessRegressor(kernel=kernel, n_restarts_optimizer=0)
+                ymodel_stack = np.zeros((self.Ntimes, self.Npols))
```
plaplant: It seems like this variable is no longer used. Can we just remove it?

jsdillon: You're right. I removed it.

hera_qm/tests/test_firstcal_metrics.py
```diff
@@ -229,7 +229,7 @@ def test_run_metrics_two_pols(self):
         # but you never know.
         two_pol_known_results = os.path.join(DATA_PATH, 'example_two_polarization_firstcal_results.hdf5')
         np.random.seed(0)
-        self.FC.run_metrics(std_cut=1.0)
+        self.FC.run_metrics(std_cut=.5)
```
adampbeardsley: Is there a lot of flagged data in the input file here? Or should we be concerned this keyword had to change significantly for the known_output to work?

mkolopanis: To be fair the "known_input" was acquired by calling running the metrics by hand with fixed environment variables (like the seed though no real random things were happening). The version of the code produced the same results on a single polarization test file we had before. The fact that it is different now might just mean we should update the test file, but log exactly what it is that is changing.

jsdillon:
@adampbeardsley I updated the example .hdf5 file, which is equivalent to updating the unit tests.

hera_qm/tests/test_firstcal_metrics.py
jsdillon: I changed the `std_cut` only because it matches the the default value and when I reran it from the command line to make the new example file that's what I used. I can easily put it back. It's arbitrary.

So, the original input data file was 100% flagged, which made no sense. I don't know how that happened. I modified it so that it was only flagged where the gains were identically equal to 1. 

Before I updated the two_pol_known_results, I tried running the unit tests using the previous example file and found that it only disagreed on the fields that I would expect disagreement given a proper accounting of flags.

mkolopanis: Thanks Josh. That is reassuring for me at least. I'll let @plaplant or @adampbeardsley weigh in for finality.

PR #276, opened 2019-05-07
Remove uvflag

adampbeardsley:
Hey @mwilensky768 would you mind reviewing this to finish off the porting we've been working on?


hera_qm/tests/test_xrfi.py
```diff
@@ -17,7 +17,7 @@
 from pyuvdata import utils as uvutils
```
mwilensky768: I don't think this import is actually used.

PR #277, opened 2019-05-09
Remove single letter variables

adampbeardsley:
I've gone through the modules and scripts and removes (almost) all single letter variables. Exceptions are when the variable is only used in one line (list comprehensions), and a couple instances where it's a single line for loop.

I have not gone through the test code.

### DIFF
--- a/hera_qm/firstcal_metrics.py
+++ b/hera_qm/firstcal_metrics.py
@@ -135,7 +135,8 @@ def plot_stds(metrics, fname=None, ax=None, xaxis='ant', kwargs={}, save=False):
         ax.set_xlim(-1, Nants)
         ax.set_xticks(range(Nants))
         ax.set_xticklabels(metrics['ants'])
-        [label.set_rotation(20) for label in ax.get_xticklabels()]
+        for label in ax.get_xticklabels():
+            label.set_rotation(20)
         ax.set_xlabel('antenna number', fontsize=14)
         ax.set_ylabel('delay solution standard deviation [ns]', fontsize=14)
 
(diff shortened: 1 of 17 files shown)


hera_qm/firstcal_metrics.py
```diff
@@ -135,7 +135,7 @@ def plot_stds(metrics, fname=None, ax=None, xaxis='ant', kwargs={}, save=False):
         ax.set_xlim(-1, Nants)
         ax.set_xticks(range(Nants))
         ax.set_xticklabels(metrics['ants'])
-        [t.set_rotation(20) for t in ax.get_xticklabels()]
+        [label.set_rotation(20) for label in ax.get_xticklabels()]
```
mkolopanis: Curious this list comprehension was left. Can we change this to a for loop to match the plotting code in omnical_metrics.

hera_qm/firstcal_metrics.py
```diff
@@ -145,7 +145,7 @@ def plot_stds(metrics, fname=None, ax=None, xaxis='ant', kwargs={}, save=False):
         ax.grid(True, zorder=0)
         ax.tick_params(size=8)
         ax.plot(xax, yax, c='k', marker='.', linestyle='-', alpha=0.85, zorder=1)
-        [t.set_rotation(20) for t in ax.get_xticklabels()]
+        [label.set_rotation(20) for label in ax.get_xticklabels()]
```
mkolopanis: also this comprehension to a loop

hera_qm/firstcal_metrics.py
adampbeardsley: Yeah, I mostly got tired. I'll fix it

PR #288, opened 2019-07-09
fix installation order issues

steven-murray:
Should fix the installation order issue in the same way that `pyuvdata` did recently.

Also adds some installation instructions and an `environment.yml` for simpler installation of dev environment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ install:
   - conda info -a
 
   # create environment and install dependencies
-  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION pip coverage
+  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION coveralls
   - source activate test-environment
   - conda install numpy scipy matplotlib astropy scikit-learn h5py coveralls aipy six pycodestyle pydocstyle pytest-cov
   - pip install git+https://github.com/RadioAstronomySoftwareGroup/pyuvdata.git
--- a/environment.yml
+++ b/environment.yml
@@ -13,6 +13,7 @@ dependencies:
   - scipy
   - pytest
   - scikit-learn # from hera_cal
+  - pytest-cov
   - pip:
     - git+https://github.com/HERA-Team/hera_cal
 
--- a/environment.yml
+++ b/environment.yml
@@ -1,0 +1,19 @@
+name: linsolve
+channels:
+  - defaults
+  - conda-forge
+dependencies:
+  - nose
+  - numpy>=1.10
+  - pip
+  - astropy>=2.0
+  - h5py
+  - six
+  - pyuvdata
+  - scipy
+  - pytest
+  - scikit-learn # from hera_cal
+  - pip:
+    - git+https://github.com/HERA-Team/hera_cal
+
+
--- a/environment.yml
+++ b/environment.yml
@@ -1,4 +1,4 @@
-name: linsolve
+name: hera_qm
 channels:
   - defaults
   - conda-forge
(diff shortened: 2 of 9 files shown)


.travis.yml
```diff
@@ -28,13 +28,11 @@ install:
...
   # create environment and install dependencies
-  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION pip coverage
+  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION coveralls
   - source activate test-environment
-  - conda install numpy scipy matplotlib astropy scikit-learn h5py coveralls aipy six pycodestyle pydocstyle pytest-cov
-  - pip install git+https://github.com/RadioAstronomySoftwareGroup/pyuvdata.git
-  - pip install git+https://github.com/HERA-Team/linsolve.git
-  - pip install git+https://github.com/HERA-Team/uvtools.git
-  - pip install git+https://github.com/HERA-Team/hera_cal.git
+  - pip install git+https://github.com/HERA-Team/linsolve.git   # only required until it updates
```
bhazelton: @steven-murray what does this comment mean?

steven-murray: It means that once the upstream PRs have been merged (eg linsolve and uvtools), these commands will not be necessary -- they can go straight into the `environment.yml`. I just wanted to remind ourselves of this so they don't stay there redundantly :-)

bhazelton: ok, makes sense. But I wouldn't take from this comment that they need to be moved somewhere else as opposed to just removed.

steven-murray: You're right, I just took them out. This PR isn't going to be merged until the upstream PRs are merged anyway.

environment.yml
```diff
@@ -0,0 +1,19 @@
+name: linsolve
```
mkolopanis: This might be nit-picky but I would appreciate if the name was the same as the environment being created in conda.

environment.yml
```diff
@@ -0,0 +1,19 @@
...
+  - pip
+  - astropy>=2.0
+  - h5py
+  - six
+  - pyuvdata
+  - scipy
+  - pytest
+  - scikit-learn # from hera_cal
+  - pip:
+    - git+https://github.com/HERA-Team/hera_cal
```
mkolopanis: `hera_cal` requires aipy and as a result the installation is failing on travis. I think you'll probably just have to add aipy back in here or wait for a `hera_cal` which make aipy auto-downloaded as a dependency through pip.

mkolopanis: also please add `pytest-cov` it is necessary for generating coverage reports.

.travis.yml
```diff
@@ -28,13 +28,9 @@ install:
   - conda info -a
 
   # create environment and install dependencies
-  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION pip coverage
+  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION coveralls
```
mkolopanis: Is there a particular reason coveralls installed here and not in the yml file like all the other modules?

environment.yml
steven-murray: Woops. Fixed in 2b82e0daa1652123ebe93d842f422875cb27cefb

environment.yml
steven-murray: Added pytest-cov. I'll wait for the upstream fixes for `aipy`, rather than putting redundant stuff in these commits. ATM, `hera_cal` is waiting on `pyuvsim`... so yeah, a bit of a chain here haha.

.travis.yml
steven-murray: My reasoning was that `coveralls` is not required on a developer's environment, *only* on travis (as it just sends the info from travis to coveralls.io)

mkolopanis: hm okay fair enough

PR #293, opened 2019-07-15
Take out unnecessary data read

### LINKED ISSUE #66: Remove extra file read in ant_metrics_run
In our quick solution to get things working, we read a data file in ant_metrics_run in order to build the antenna array. We should refactor the inputs of Antenna_Metrics.__init__() so files are either read outside, or building the aa is done inside.

adampbeardsley:
Closes #66


adampbeardsley:
Crap, this isn't working yet. I tried it on a uvh5 file, but the metaread works differently on miriad, which is what is tested.

adampbeardsley:
I think this PR is ready for review.

PR #296, opened 2019-07-19
Update XRFI Day Thresholding

jsdillon:
Modifies xrfi day-thresholding (which flags entire channels and entire times based on z-scores of median metrics over the other dimension) in three ways:

1. Removes detrending of 1D median metrics when computing modified z-scores from median-ed metrics. This detrending was creating problems where times that were almost entirely flagged by the original metrics waterfalls didn't trigger the threshold because the median metric of neighboring times were noisy, suppressing the apparent outlierness. This was discussed on the 7/18/19 QM telecon.
2. Adds a 1D watershed algorithm to thresholding. Now moderately bad channels/integrations next to very bad channels/integrations also get flagged.
3. Bumps up the default thresholds to 7 sigma and 3 sigma for neighbors in the watershed.

### DIFF
--- a/hera_qm/xrfi.py
+++ b/hera_qm/xrfi.py
@@ -1492,8 +1492,8 @@ def xrfi_run(ocalfits_file, acalfits_file, model_file, data_file, history,
         uvf.write(outpath, clobber=clobber)
 
 
-def day_threshold_run(data_files, history, kt_size=8, kf_size=8, nsig_f=5.0, nsig_t=5.0,
-                      clobber=False):
+def day_threshold_run(data_files, history, nsig_f=7., nsig_t=7.,
+                      nsig_f_adj=3., nsig_t_adj=3., clobber=False):
     """Apply thresholding across all times/frequencies, using a full day of data.
 
     This function will write UVFlag files for each data input (omnical gains,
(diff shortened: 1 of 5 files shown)


hera_qm/xrfi.py
```diff
@@ -679,6 +679,38 @@ def watershed_flag(uvf_m, uvf_f, nsig_p=2., nsig_f=None, nsig_t=None, avg_method
     return uvf
 
 
+def _ws_flag_1D(metric, fin, nsig=2.):
```
adampbeardsley: I actually just noticed the `_ws_flag_waterfall` function does take 1D arrays. Perhaps we should replace the code there with your code instead of making a redundant function?

adampbeardsley: On second thought... your code does something slightly different -- it does not iterate until the region has stopped growing. Was this intentional? And should we rename some functions to make it more clear?
Maybe we still put your code in the original function, but add an option to not iterate?

jsdillon: I missed the while statement in the original ws_flagging. I've harmonized the two into a single, more elegant function.

jsdillon:
And for the record, he's non-iterative watershed: 
non_iterative_watershed

vs. iterative watershed:
iterative_watershed

And here's the difference between the two
iterative_vs_non_iterative

jsdillon:
And here's some plots of the effects on the flagging fraction vs. time / freq:

Screen Shot 2019-07-19 at 3 30 46 PM
Screen Shot 2019-07-19 at 3 30 30 PM

PR #310, opened 2019-10-30
X orientation update

adampbeardsley:
A couple fixes to properly handle x_orientation. But probably not complete, so this is still a draft.


mkolopanis:
I think that maybe there is more to do, but it needs some fresh eyes a la review style.

mkolopanis:
I would also like to add that RTP on site currently Runs on code from this branch back at commit https://github.com/HERA-Team/hera_qm/pull/310/commits/c5360373ddd7d795e3bfce0121eeba921ae64e0a. I think if people can find time, it would make me feel better if RTP did not run on random hot_fixes in a branch.

Issue #313, opened 2019-11-27
End python 2.7 Support


mkolopanis:
python 2.7 end of life is 1 Jan 2020. We should prepare for this by ending active support.

adampbeardsley:
Do we actively support? Or does this just mean turning off the CI testing?

mkolopanis:
I would argue turning off tests, and bending over backwards to support it with `six` stuff and all the "trying to make bytes vs unicode" junk go away. I'm sure @plaplant  has many thoughts on the matter

mkolopanis:
Should we tag a version for final python 2 compatible version?

plaplant:
Yeah, we did something similar on `hera_cal`: https://github.com/HERA-Team/hera_cal/tree/last-py2-compatible. I guess it's a branch instead of a tag, I have a slight preference for it being a tag.

mkolopanis:
I do as well.

mkolopanis:
this was done with #314

PR #314, opened 2019-12-06
remove python 2

mkolopanis:
addresses #313


mkolopanis:
probably want a version/tag/release before this to mark final python 2 support maybe after too to bump to a new version

PR #320, opened 2020-04-07
GitHub actions

adampbeardsley:
Try setting up github actions for continuous integration


.github/workflows/linux_ci.yml
```diff
@@ -0,0 +1,54 @@
+name: Run Tests
+
+on:
+  pull_request:
+    branches: [ master ]
```
mkolopanis, line 5: So this will only execute Actions on a pull request to be merged into `master` and only on the "merged" head (like travis's push/pr builds but just the pr one). If that is the desired functionality you're good to go. If instead you want to use this as your main testing base for all pushes on branches and PR's you may want to change to:
```
on: [push, pull_request]
```

.github/workflows/linux_ci.yml
```diff
@@ -0,0 +1,54 @@
...
+        source activate ${ENV_NAME}
+        pip install .
+
+    - name: Run Tests
+      run: |
+        export PATH="$HOME/miniconda/bin:$PATH"
+        source activate $ENV_NAME
+        pytest --cov-report xml:./coverage.xml
+        pycodestyle
+        pydocstyle hera_qm/*.py --convention=numpy
```
mkolopanis, line 49: While there's no reason you can't leave these as one giant part of an Action but you can also split them up into separate jobs entirely if you wan. Either way up to you!

.github/workflows/linux_ci.yml
```diff
@@ -0,0 +1,54 @@
...
+  tests:
+    name: Ubuntu Tests
+    env:
+      ENV_NAME: hera_qm_tests
+      PYTHON: ${{ matrix.python-version }}
+      OS: ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
```
mkolopanis, line 17: if you're feeling feisty you can enable macos-latest as an OS, but you only get 5 concurrent mac builds at a time. and well mac can be testy some times.

ci/hera_qm_tests.yml
```diff
@@ -0,0 +1,21 @@
+name: hera_qm_tests
+channels:
+  - defaults
+  - conda-forge
+dependencies:
+  - numpy>=1.10
+  - pip
+  - astropy>=2.0.16
```
mkolopanis, line 8: Since you're only testing in python 3 you might want to change this to be `astropy>=3.2.3`, that was when the UNSO changes were.

adampbeardsley:
The only thing I'd like to do at this point is define "passing" for codecov... I can't seem to find it in the docs anywhere.

.github/workflows/linux_ci.yml
adampbeardsley: Yeah, I kinda figured we don't need to run on every push. Let's leave it here for now, and we can switch if people want the tests run more frequently.

.github/workflows/linux_ci.yml
adampbeardsley: Let's give it a try!
I don't think many people are running qm on macs (mostly on nrao/enterprise/on site), but might as well keep track of it. We can always run it but not make it required.

ci/hera_qm_tests.yml
adampbeardsley: Will do.

PR #324, opened 2020-04-23
Add option to exclude autocorrelations from raw data meanfilt step in xrfi_run

### LINKED ISSUE #323: Add option to flag only on crosses
https://github.com/HERA-Team/hera_qm/blob/3f2d0a66a45384394e9e00693a23da652df2b422/hera_qm/xrfi.py#L1528

H3C seems to show persistent structures in the autos known as "icicles": https://github.com/HERA-Team/HERA_Commissioning/issues/509

It's unclear if they affect the cross-correlations, but if they don't, we don't want to use the autos for flagging.

jsdillon:
This closes #323

### DIFF
--- a/hera_qm/xrfi.py
+++ b/hera_qm/xrfi.py
@@ -1374,7 +1374,7 @@ def chi_sq_pipe(uv, alg='zscore_full_array', modified=False, sig_init=6.0,
 
 def xrfi_run(ocalfits_file, acalfits_file, model_file, data_file, history,
              xrfi_path='', kt_size=8, kf_size=8, sig_init=5.0, sig_adj=2.0,
-             ex_ants=None, exclude_autos=False, metrics_file=None, clobber=False,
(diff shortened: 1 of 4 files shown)


hera_qm/xrfi.py
```diff
@@ -1526,6 +1528,8 @@ def xrfi_run(ocalfits_file, acalfits_file, model_file, data_file, history,
     # Read in data file
     uv_d = UVData()
     uv_d.read(data_file)
+    if exclude_autos:
+        uv_d.select(ant_str='cross')
```
adampbeardsley, line 1532: Hmm, would it be even better to pass in `ant_str`? Then we could do:
```suggestion
    uv_d.read(data_file, ant_str=ant_str)
```

jsdillon: Done.

jsdillon:
I believe your comments have been addressed @adampbeardsley

PR #335, opened 2020-06-23
xrfi chunking

adampbeardsley:
This PR coincides with the opm file chunking branch to enable working in batches for XRFI.

The new wrapper takes in several data files, performs the xrfi algorithm, and stores outputs for the inner parts of the chunk - discarding data from edge files. On the OPM side, the do script will assign overlapping chunks of files to jobs so in the end we get output from all files. 

Because we are discarding the edges, we don't need to pad the data anymore. I've added an option to the algorithm functions to turn off padding, which will save a little time on the median filter.

I've run this through the pipeline on NRAO, but still need to check the outputs. But now's a good time to start getting more eyes on the code.


hera_qm/xrfi.py
```diff
@@ -389,6 +389,9 @@ def detrend_medfilt(data, flags=None, Kt=8, Kf=8):
     Kf : int, optional
         The box size in frequency (second) dimension to apply medfilt over. Default
         is 8 pixels.
+    pad : bool, optional
+        If True (default), pad the data by mirroring at the edges by the size of
```
jsdillon, line 393: Clarify that this is in both dimensions.

hera_qm/xrfi.py
```diff
@@ -1639,6 +1658,310 @@ def xrfi_run(ocalfits_file, acalfits_file, model_file, data_file, history,
         uvf.write(outpath, clobber=clobber)
 
 
+def xrfi_h3c_idr2_1_run(ocalfits_files, acalfits_files, model_files, data_files,
```
jsdillon, line 1661: Again, I'd consider renaming this to something a bit more descriptive.

hera_qm/xrfi.py
```diff
@@ -1639,6 +1658,310 @@ def xrfi_run(ocalfits_file, acalfits_file, model_file, data_file, history,
...
+    run_check_acceptability : bool
+        Option to check acceptable range of the values of parameters
+        on UVFlag Object.
+
+    Returns
+    -------
+    None
+
+    """
+    history = 'Flagging command: "' + history + '", Using ' + hera_qm_version_str
```
jsdillon, line 1729: It's slightly weird that whatever you put in the "history" is labeled as a flagging command, even if that's how it's primarily used. Maybe change the argument name?

hera_qm/xrfi.py
```diff
@@ -1639,6 +1658,310 @@ def xrfi_run(ocalfits_file, acalfits_file, model_file, data_file, history,
...
+
+    # Initial run on cal data products
+    # Calculate metric on abscal data
+    uvc_a = UVCal()
+    uvc_a.read_calfits(acalfits_files)
+    uvf_apriori = UVFlag(uvc_a, mode='flag', copy_flags=True, label='A priori flags.')
+    uvf_ag, uvf_agf = xrfi_pipe(uvc_a, alg='detrend_medfilt', Kt=kt_size, Kf=kf_size, xants=xants,
+                                cal_mode='gain', sig_init=sig_init, sig_adj=sig_adj,
+                                label='Abscal gains, round 1.', pad=False,
+                                run_check=run_check,
```
jsdillon, line 1746: If you define checks = {'run_check': run_check} etc. than you can just ** these kwargs in and save some lines.

jsdillon: Or go broader: make every argument that never changes get passed via a single dict.

hera_qm/xrfi.py
```diff
@@ -1639,6 +1658,310 @@ def xrfi_run(ocalfits_file, acalfits_file, model_file, data_file, history,
...
+
+    # Calculate metric on data file
+    uvf_d2, uvf_df2 = xrfi_pipe(uv_d, alg='detrend_meanfilt', xants=[], Kt=kt_size, Kf=kf_size,
+                                sig_init=sig_init, sig_adj=sig_adj,
+                                label='Data, round 2.', pad=False,
+                                run_check=run_check,
+                                check_extra=check_extra,
+                                run_check_acceptability=run_check_acceptability)
+
+    # Get the absolute chi-squared values
```
jsdillon, line 1877: Mislabeled, it's omnical.

hera_qm/xrfi.py
```diff
@@ -1639,6 +1658,310 @@ def xrfi_run(ocalfits_file, acalfits_file, model_file, data_file, history,
...
+    uv_v.read(model_files, axis='blt')
+    uvf_v, uvf_vf = xrfi_pipe(uv_v, alg='detrend_medfilt', xants=[], Kt=kt_size, Kf=kf_size,
+                              sig_init=sig_init, sig_adj=sig_adj,
+                              label='Omnical visibility solutions, round 1.',
+                              pad=False,
+                              run_check=run_check,
+                              check_extra=check_extra,
+                              run_check_acceptability=run_check_acceptability)
+
+    # Get the absolute chi-squared values
```
jsdillon, line 1786: mislabeled, or confusing at least. It's omnical, not abscal

hera_qm/xrfi.py
jsdillon: Or maybe just confusing

hera_qm/xrfi.py
```diff
@@ -1639,6 +1658,310 @@ def xrfi_run(ocalfits_file, acalfits_file, model_file, data_file, history,
...
+    # Determine the actual files to store
+    uvtemp = UVData()
+    uvtemp.read(data_files[0], read_data=False)
+    nintegrations = len(data_files) * uvtemp.Ntimes
+    ndrop = int(np.ceil(kt_size / uvtemp.Ntimes))
+    start_ind = ndrop
+    end_ind = len(data_files) - ndrop
+    # If we're the first or last job, store all flags for the edge
+    datadir = os.path.dirname(os.path.abspath(data_files[0]))
+    all_files = sorted(glob.glob(os.path.join(datadir, '*sum.uvh5')))
```
jsdillon, line 1938: Not wild about hardcoding this...

jsdillon: Can we have it figure out the files based on the data files passed in.

hera_qm/xrfi.py
```diff
@@ -1639,6 +1658,310 @@ def xrfi_run(ocalfits_file, acalfits_file, model_file, data_file, history,
...
+                'ax_metrics2.h5': uvf_ax2, 'ax_flags2.h5': uvf_axf2,
+                'data_metrics2.h5': uvf_d2, 'data_flags2.h5': uvf_df2,
+                'chi_sq_renormed2.h5': uvf_chisq2, 'chi_sq_flags2.h5': uvf_chisq_f2,
+                'combined_metrics2.h5': uvf_metrics2, 'combined_flags2.h5': uvf_fws2,
+                'flags2.h5': uvf_combined2}
+
+    # Determine the actual files to store
+    uvtemp = UVData()
+    uvtemp.read(data_files[0], read_data=False)
+    nintegrations = len(data_files) * uvtemp.Ntimes
```
jsdillon, line 1932: Can we assume that all files have the same number of integrations?

hera_qm/xrfi.py
```diff
@@ -1639,6 +1658,310 @@ def xrfi_run(ocalfits_file, acalfits_file, model_file, data_file, history,
...
+    Several files are concatenated together to perform the detrending,
+    and the flags from the inner files are stored.
+    The algorithm is run on several data products: omnical gains, omnical chisq,
+    abscal gains, abscal chisq, omnical visibility solutions, renormalized chisq,
+    and the raw data. All of these, except the data, are run twice - first to
+    get an initial estimate of heavily contaminated data, and a second time
+    to get better estimate. The metrics and flags from each data product and both
+    rounds are stored in the xrfi_path (which defaults to a subdirectory, see
+    xrfi_path below). Also stored are the a priori flags and combined metrics/flags.
+
```
jsdillon, line 1678: Explain which input files actually get saved and how the beginning and end of the list are handled.

hera_qm/xrfi.py
```diff
@@ -1639,6 +1658,310 @@ def xrfi_run(ocalfits_file, acalfits_file, model_file, data_file, history,
...
+                'og_metrics2.h5': uvf_og2, 'og_flags2.h5': uvf_ogf2,
+                'ox_metrics2.h5': uvf_ox2, 'ox_flags2.h5': uvf_oxf2,
+                'ag_metrics2.h5': uvf_ag2, 'ag_flags2.h5': uvf_agf2,
+                'ax_metrics2.h5': uvf_ax2, 'ax_flags2.h5': uvf_axf2,
+                'data_metrics2.h5': uvf_d2, 'data_flags2.h5': uvf_df2,
+                'chi_sq_renormed2.h5': uvf_chisq2, 'chi_sq_flags2.h5': uvf_chisq_f2,
+                'combined_metrics2.h5': uvf_metrics2, 'combined_flags2.h5': uvf_fws2,
+                'flags2.h5': uvf_combined2}
+
+    # Determine the actual files to store
```
jsdillon, line 1929: I find this whole procedure a little confusing. You definitely want to comment better what's going on... Would it make more sense for the user to specify which file JD strings are to be completely flagged, RFI flagged, or not written out at all?

hera_qm/xrfi.py
adampbeardsley: I don't think so. This is just used to determine the edges of the entire night, which is usually outside the files that are passed in. I've updated the search to be more robust.

hera_qm/xrfi.py
adampbeardsley: For a given day, yes. I think it would require a correlator restart to have different integrations/file. This logic will probably have to be revisited with BDA... I'm not sure how `Ntimes` is defined there.

adampbeardsley:
@jsdillon things got a little disorganized, but I think I addressed all your comments... except the one about function name which I'd like to punt down the road to the restructuring.
Let me know if I missed anything.

adampbeardsley:
@jsdillon (or @plaplant ) could I get you to approve again? I was still missing test coverage :(

PR #336, opened 2020-06-29
Option to enforce separable flags during day thresholding. 

anguta:
Broadcast flags based on the percentage of flags in time and frequency.

### DIFF
--- a/hera_qm/utils.py
+++ b/hera_qm/utils.py
@@ -294,6 +294,12 @@ def get_metrics_ArgumentParser(method_name):
         ap.add_argument("--run_if_first", default=None, type=str, help='only run \
                         day_threshold_run if the first item in the sorted data_files \
                         list matches run_if_first (default None means always run)')
+        ap.add_argument("--separable_flags", default=False, action="store_true", type=bool,
+                        help="Generate separable flags by thresholding in time and frequency." )
+        ap.add_argument("--time_threshold", default=0.15, type=float,
+                        help="Fully flag times with flagging fraction over this threshold. Default is 0.15.")
+        ap.add_argument("--freq_threshold", default=0.02, type=float,
+                        help="Fully flag frequencies with flagging fraction over this threshold. Default is 0.02.")
     elif method_name == 'xrfi_apply':
         ap.prog = 'xrfi_apply.py'
         ap.add_argument('--infile_format', default='miriad', type=str,
--- a/hera_qm/utils.py
+++ b/hera_qm/utils.py
@@ -294,8 +294,8 @@ def get_metrics_ArgumentParser(method_name):
         ap.add_argument("--run_if_first", default=None, type=str, help='only run \
                         day_threshold_run if the first item in the sorted data_files \
                         list matches run_if_first (default None means always run)')
-        ap.add_argument("--separable_flags", action="store_true", default=False,
-                        help="Generate separable flags by thresholding in time and frequency." )
+        ap.add_argument("--broadcast_flags", action="store_true", default=False,
+                        help="Broadcast final day-thresholded flags in time and frequency." )
         ap.add_argument("--time_threshold", default=0.15, type=float,
                         help="Fully flag times with flagging fraction over this threshold. Default is 0.15.")
         ap.add_argument("--freq_threshold", default=0.02, type=float,
--- a/hera_qm/utils.py
+++ b/hera_qm/utils.py
@@ -294,6 +294,8 @@ def get_metrics_ArgumentParser(method_name):
         ap.add_argument("--run_if_first", default=None, type=str, help='only run \
                         day_threshold_run if the first item in the sorted data_files \
                         list matches run_if_first (default None means always run)')
+        ap.add_argument("--separable_flags", action="store_true", default=False,
+                        help="Do not OR thresholded flags with previous flags.")
         ap.add_argument("--broadcast_flags", action="store_true", default=False,
                         help="Broadcast final day-thresholded flags in time and frequency." )
         ap.add_argument("--time_threshold", default=0.15, type=float,
--- a/hera_qm/utils.py
+++ b/hera_qm/utils.py
@@ -294,7 +294,7 @@ def get_metrics_ArgumentParser(method_name):
         ap.add_argument("--run_if_first", default=None, type=str, help='only run \
                         day_threshold_run if the first item in the sorted data_files \
                         list matches run_if_first (default None means always run)')
-        ap.add_argument("--separable_flags", default=False, action="store_true", type=bool,
+        ap.add_argument("--separable_flags", action="store_true", default=False,
                         help="Generate separable flags by thresholding in time and frequency." )
         ap.add_argument("--time_threshold", default=0.15, type=float,
                         help="Fully flag times with flagging fraction over this threshold. Default is 0.15.")
--- a/hera_qm/xrfi.py
+++ b/hera_qm/xrfi.py
@@ -1683,6 +1683,8 @@ def day_threshold_run(data_files, history, nsig_f=7., nsig_t=7.,
         Option to check acceptable range of the values of parameters
         on UVFlag Object.
     separable_flags : bool, optional
+        !!! WARNING -- THIS FEATURE IS EXPERIMENTAL AND DEFAULTS ARE NOT !!!
+        !!! CALIBRATED TO PRODUCE ANY SORT OF REASONABLE RESULT !!!
         If set to True, then force the flags to be separable by performing
         a final thresholding for each channel and time by flagging times /
         channels if fraction of flags in each time is above time_threshold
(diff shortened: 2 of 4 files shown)


hera_qm/xrfi.py
```diff
@@ -1642,7 +1642,8 @@ def xrfi_run(ocalfits_file, acalfits_file, model_file, data_file, history,
 def day_threshold_run(data_files, history, nsig_f=7., nsig_t=7.,
                       nsig_f_adj=3., nsig_t_adj=3., clobber=False,
                       run_check=True, check_extra=True,
-                      run_check_acceptability=True):
+                      run_check_acceptability=True, separable_flags=False,
+                      time_threshold=0.15, freq_threshold=0.01):
```
adampbeardsley, line 1646: It's minor, but could you put the new parameters in front of the check keywords, for consistency across the module?

hera_qm/utils.py
```diff
@@ -294,6 +294,12 @@ def get_metrics_ArgumentParser(method_name):
         ap.add_argument("--run_if_first", default=None, type=str, help='only run \
                         day_threshold_run if the first item in the sorted data_files \
                         list matches run_if_first (default None means always run)')
+        ap.add_argument("--separable_flags", action="store_true", default=False,
+                        help="Generate separable flags by thresholding in time and frequency." )
+        ap.add_argument("--time_threshold", default=0.15, type=float,
+                        help="Fully flag times with flagging fraction over this threshold. Default is 0.15.")
+        ap.add_argument("--freq_threshold", default=0.02, type=float,
+                        help="Fully flag frequencies with flagging fraction over this threshold. Default is 0.02.")
```
adampbeardsley, line 302: Default in the function itself is 0.01. Maybe make consistent?

hera_qm/xrfi.py
```diff
@@ -1732,6 +1751,21 @@ def day_threshold_run(data_files, history, nsig_f=7., nsig_t=7.,
...
     uvf_total |= UVFlag(files)
 
+    # Generate Separable Flags.
+    if separable_flags:
+        broadcast_flags = np.zeros(uvf_total.flag_array.shape).astype(bool)
+        for pol in range(uvf_total.flag_array.shape[2]):
+            fa = uvf_total.flag_array[:, :, pol]
+            # First, flag all frequencies above threshold.
+            for m in range(fa.shape[1]):
+                if np.count_nonzero(~fa[:, m]) / fa.shape[0] >= freq_threshold:
```
adampbeardsley, line 1761: Why is `fa` negated? This looks like you're counting the number of non-flagged elements?

anguta:
> I have a couple small comments, and a more significant one:
> Have you looked at how this actually works out on data? In particular, this is being done after some time/thresholding has already happened, and if there's a significant amount there, you can easily end up flagging the entire observation. For example, if 1% of times are completely flagged, you will end up flagging everything.

Thanks for the review! 

I have not studied this on actual data yet so the default values may not be very well calibrated. What I do know is that foreground and xtalk filtering (linear and I think for similar reasons CLEAN) do not commute unless the flags are constant in time (when the entire time is not flagged). I did have a sense that some thresholding was being done but it didn't seem to produce strictly separable flags when I tried it (perhaps incorrectly). With all this said, this feature is basically experimental and being added to let me investigate this issue. Without well established default params, perhaps this is not ready for merging yet. On the other hand, I don't like to leave PRs hanging since i keep having to merge over and over again. Perhaps the right approach here is to put in warnings that this feature is experimental and that the defaults are not calibrated to be reasonable in real world situations. What do you think? 

Do you think it makes more sense to integrate this sort of functionality into the lines of code where this previous type of thresholding is being done? If so, where should I look?

cheers,

-Aaron

hera_qm/utils.py
anguta: Will do.

hera_qm/xrfi.py
anguta: Thanks for catching this bug!

adampbeardsley:
I see. Yeah, I suspect it would be better to incorporate this into the original thresholding rather than a separate step. I'm not sure this will be exactly what you want to do, but you could imagine replacing the lines:
```
    # Read non thresholded flags and combine
    files = [glob.glob(d + '/*.flags2.h5')[0] for d in xrfi_dirs]
    uvf_total |= UVFlag(files)
```
With:
```
    if not separable_flags:
        # Read non thresholded flags and combine
        files = [glob.glob(d + '/*.flags2.h5')[0] for d in xrfi_dirs]
        uvf_total |= UVFlag(files)
```
Up to that point, `uvf_total` is a bunch of thresholded waterfalls OR'd together, so it actually is separable.

Also, have you looked into the `hera_pspec` version? I can't remember the name of it, but they have a function to do what you want... We should either use that, or somehow synergize so we don't have too much repeated code out there. I can dig more if you need me to.

anguta:
> I see. Yeah, I suspect it would be better to incorporate this into the original thresholding rather than a separate step. I'm not sure this will be exactly what you want to do, but you could imagine replacing the lines:
> 
> ```
>     # Read non thresholded flags and combine
>     files = [glob.glob(d + '/*.flags2.h5')[0] for d in xrfi_dirs]
>     uvf_total |= UVFlag(files)
> ```
> 
> With:
> 
> ```
>     if not separable_flags:
>         # Read non thresholded flags and combine
>         files = [glob.glob(d + '/*.flags2.h5')[0] for d in xrfi_dirs]
>         uvf_total |= UVFlag(files)
> ```
> 
> Up to that point, `uvf_total` is a bunch of thresholded waterfalls OR'd together, so it actually is separable.
> 

I guess I wasn't super clear about what I mean by `separable`. I mean by `separable` that the inverse of the flags can be expressed as the outer product of a time and frequency vector that have zeros on flagged integrations/channels and ones elsewhere so I think this is separable in a different sense then what you are referring to here. 

> Also, have you looked into the `hera_pspec` version? I can't remember the name of it, but they have a function to do what you want... We should either use that, or somehow synergize so we don't have too much repeated code out there. I can dig more if you need me to.

Re `hera_pspec`. The function that I am aware of is `broadcast_dset_flags`. Unfortunately, this function is a property of the `pspecdata` object so I don't think it can easily be deployed for these purposes. Ideally, think that this function belongs in `hera_qm` and would be called by `hera_pspec` so maybe it would make sense to break this code out into a utility function that is called by `xrfi_day_threshold` here and is available for `hera_pspec` to call.

adampbeardsley:
Ok, I agree using pspec for this would be tedious. If we do end up with a simple thresholding like your initial PR, we may want to look into making a function in uvtools or somewhere that both repos can access without adding dependencies. Maybe that's not possible.

But I do think we are thinking of separable in the same fashion. `uvf_total` is made by taking a bunch of input UVFlag objects and identifying problematic times and channels. The only flags in it, prior to the snippet I pasted in, are made by flagging entire channels and entire integrations. The snippet I pasted ORs that with the regular xrfi flags (thresholded pixels + watershed).
My proposal there would not catch bright blips, only stuff that's persistent across time/frequency. But I think yours would do the same?

Either way you proceed, I think it's worth trying it out with at least some test data to see how it looks before merging.

anguta:
> Ok, I agree using pspec for this would be tedious. If we do end up with a simple thresholding like your initial PR, we may want to look into making a function in uvtools or somewhere that both repos can access without adding dependencies. Maybe that's not possible.
> 
> But I do think we are thinking of separable in the same fashion. `uvf_total` is made by taking a bunch of input UVFlag objects and identifying problematic times and channels. The only flags in it, prior to the snippet I pasted in, are made by flagging entire channels and entire integrations. The snippet I pasted ORs that with the regular xrfi flags (thresholded pixels + watershed).
> My proposal there would not catch bright blips, only stuff that's persistent across time/frequency. But I think yours would do the same?
> 
> Either way you proceed, I think it's worth trying it out with at least some test data to see how it looks before merging.

 I think I now see what you are saying. The ''thresholded flags'' are separable and then they are or'd with the non separable flags from the various metrics. It seems like the thresholds are currently being determined by metrics (rather then flags) so the functionality is a bit different from what I want exactly but in principal, the separable flags could be achieved by forgoing the ORing step and just using the thresholding flags. 

Since I would like to do a final percent flags threshold similar to what is done in `hera_pspec`, I think it makes sense to add some new code that does this specific type of thresholding. I've broken it out into a standalone xrfi function. Putting this in `uvtools` wouldn't be too hard. It's unclear to me where in `uvtools` it belongs though. Since there is no other RFI functionality in `uvtools`, I think it makes sense in `xrfi` unless move more general `xrfi` functions into `uvtools` as well.

anguta:
@adampbeardsley, I will test this out on some data and document the results here.

anguta:
I now think there are better ways to enforce time independent flags that should be explored instead..

Issue #329, opened 2020-06-06
Make various inputs to xrfi_run optional and fix day_theshold_run accordingly


jsdillon:
Right now, `xrfi_run` is very prescriptive in exactly what inputs it needs and I think we should be more flexible.

Specifically, I think it should be able to take any of the following:
* Raw cross-correlations
* Raw auto-correlations
* Omnical chi^2 (both for local outliers and absolute outliers)
* Omnical gains
* Omnical visibility solutions
* Abscal chi^2 (both for local outliers and absolute outliers)
* Abscal gains

And I think one should be allowed to specify which steps are run using median filters and which steps are run using mean filters and which steps are run using both.

`day_threshold_run` will have to be modified to be flexible about the inputs, given the various possible outputs of `xrfi_run`.

I think this should be addressed simultaneously with #311 and #328

jsdillon:
Addressed in #338

PR #338, opened 2020-07-01
Flexible inputs

anguta:
Addresses #329

### DIFF
--- a/hera_qm/utils.py
+++ b/hera_qm/utils.py
@@ -269,8 +269,6 @@ def get_metrics_ArgumentParser(method_name):
         ap.add_argument('--ex_ants', default=None, type=str,
                         help='Comma-separated list of antennas to exclude. Flags of visibilities '
                         'formed with these antennas will be set to True.')
-        ap.add_argument('--ant_str', default=None, type=str,
-                        help='option to pass into UVData.read() setting which baselines are used for raw visibility flags')
         ap.add_argument('--metrics_file', default=None, type=str,
                         help='Metrics file that contains a list of excluded antennas. Flags of '
                         'visibilities formed with these antennas will be set to True.')
@@ -295,8 +293,6 @@ def get_metrics_ArgumentParser(method_name):
         ap.add_argument('--ex_ants', default=None, type=str,
                         help='Comma-separated list of antennas to exclude. Flags of visibilities '
                         'formed with these antennas will be set to True.')
-        ap.add_argument('--ant_str', default=None, type=str,
-                        help='option to pass into UVData.read() setting which baselines are used for raw visibility flags')
         ap.add_argument('--metrics_file', default=None, type=str,
                         help='Metrics file that contains a list of excluded antennas. Flags of '
                         'visibilities formed with these antennas will be set to True.')
@@ -357,7 +353,7 @@ def get_metrics_ArgumentParser(method_name):
                         day_threshold_run if the first item in the sorted data_files \
                         list matches run_if_first (default None means always run)')
         ap.add_argument("--skip_making_flagged_abs_calfits", default=False, action="store_true",
-                        help='If True, flag the abscal files.')
+                        help='If True, skip flagging the abscal files.')
     elif method_name == 'xrfi_apply':
         ap.prog = 'xrfi_apply.py'
         ap.add_argument('--infile_format', default='miriad', type=str,
--- a/hera_qm/utils.py
+++ b/hera_qm/utils.py
@@ -39,7 +39,7 @@ def get_metrics_ArgumentParser(method_name):
     """
     methods = ["ant_metrics", "firstcal_metrics", "omnical_metrics", "xrfi_h1c_run",
                "delay_xrfi_h1c_idr2_1_run", "xrfi_run", "xrfi_apply", "day_threshold_run",
-               "xrfi_h3c_idr2_1_run"]
+               "xrfi_h3c_idr2_1_run", "xrfi_run_data_only"]
     if method_name not in methods:
         raise AssertionError('method_name must be one of {}'.format(','.join(methods)))
 
@@ -276,6 +276,36 @@ def get_metrics_ArgumentParser(method_name):
                         'visibilities formed with these antennas will be set to True.')
         ap.add_argument("--clobber", default=False, action="store_true",
                         help='overwrites existing files (default False)')
+    elif method_name == 'xrfi_run_data_only':
+        ap.prog = 'xrfi_run_data_only.py'
+        ap.add_argument('--data_file', default=None, type=str, help='Raw visibility '
+                        'data file to flag on.')
+        ap.add_argument('--xrfi_path', default='', type=str,
+                        help='Path to save flag files to. Default is same directory as input file.')
+        ap.add_argument('--kt_size', default=8, type=int,
+                        help='Size of kernel in time dimension for detrend in xrfi '
+                        'algorithm. Default is 8.')
+        ap.add_argument('--kf_size', default=8, type=int,
+                        help='Size of kernel in frequency dimension for detrend in '
+                        'xrfi algorithm. Default is 8.')
+        ap.add_argument('--sig_init', default=6.0, type=float,
+                        help='Starting number of sigmas to flag on. Default is 6.0.')
+        ap.add_argument('--sig_adj', default=2.0, type=float,
+                        help='Number of sigmas to flag on for data adjacent to a flag. Default is 2.0.')
+        ap.add_argument('--ex_ants', default=None, type=str,
+                        help='Comma-separated list of antennas to exclude. Flags of visibilities '
+                        'formed with these antennas will be set to True.')
+        ap.add_argument('--ant_str', default=None, type=str,
+                        help='option to pass into UVData.read() setting which baselines are used for raw visibility flags')
+        ap.add_argument('--metrics_file', default=None, type=str,
+                        help='Metrics file that contains a list of excluded antennas. Flags of '
+                        'visibilities formed with these antennas will be set to True.')
+        ap.add_argument("--clobber", default=False, action="store_true",
+                        help='overwrites existing files (default False)')
+        ap.add_argument("--median_filter_cross", default=False, action="store_true",
+                        help="performs a median filter on cross-correlations. Adds significantly to runtime.")
+        ap.add_argument("--skip_mean_filter_cross", default=False, action="store_true",
+                        help="save i/o by skipping mean filter on cross correlations.")
     elif method_name == 'xrfi_h3c_idr2_1_run':
         ap.prog = 'xrfi_h3c_idr2_1_run.py'
         ap.add_argument('--ocalfits_files', nargs='+', type=str, help='Omnical '
@@ -326,6 +356,8 @@ def get_metrics_ArgumentParser(method_name):
         ap.add_argument("--run_if_first", default=None, type=str, help='only run \
                         day_threshold_run if the first item in the sorted data_files \
                         list matches run_if_first (default None means always run)')
+        ap.add_argument("--skip_making_flagged_abs_calfits", default=False, action="store_true",
+                        help='If True, flag the abscal files.')
     elif method_name == 'xrfi_apply':
         ap.prog = 'xrfi_apply.py'
         ap.add_argument('--infile_format', default='miriad', type=str,
--- a/hera_qm/xrfi.py
+++ b/hera_qm/xrfi.py
@@ -1558,9 +1558,6 @@ def xrfi_run(ocalfits_file=None, acalfits_file=None, model_file=None, data_file=
     """
     if ocalfits_file is None and acalfits_file is None and model_file is None and data_file is None:
         raise ValueError("Must provide at least one of the following; ocalfits_file, acalfits_file, model_file, data_file")
-    # if ocalfits_file is supplied so too must acalfits_file
-    if (ocalfits_file is not None and acalfits_file is None) or (ocalfits_file is None and acalfits_file is not None):
-        raise ValueError("Must provide a non-empty history string.")
     # user must provide an optional output prefix if no data file is provided.
     if output_prefix is None:
         if data_file is not None:
@@ -1824,7 +1821,7 @@ def xrfi_run(ocalfits_file=None, acalfits_file=None, model_file=None, data_file=
             flags += [uvf_axf2]
         else:
             uvf_ax2 = None; uvf_axf2 = None
-            
+
         if abscal_zscore_filter:
             uvf_az2, uvf_azf2 = chi_sq_pipe(uvc_a, alg='zscore_full_array', modified=True,
                                             sig_init=sig_init, sig_adj=sig_adj,
--- a/hera_qm/xrfi.py
+++ b/hera_qm/xrfi.py
@@ -1373,7 +1373,7 @@ def chi_sq_pipe(uv, alg='zscore_full_array', modified=False, sig_init=6.0,
 #############################################################################
 
 
-def xrfi_run(ocalfits_file=None, acalfits_file=None, model_file=None, data_file=None, history,
+def xrfi_run(ocalfits_file=None, acalfits_file=None, model_file=None, data_file=None, history=None,
              xrfi_path='', kt_size=8, kf_size=8, sig_init=5.0, sig_adj=2.0,
              ex_ants=None, ant_str=None, metrics_file=None, clobber=False,
              run_check=True, check_extra=True, run_check_acceptability=True):
@@ -1448,8 +1448,10 @@ def xrfi_run(ocalfits_file=None, acalfits_file=None, model_file=None, data_file=
     if ocalfits_file is None and acalfits_file is None and model_file is None and data_file is None:
         raise ValueError("Must provide at least one of the following; ocalfits_file, acalfits_file, model_file, data_file")
     # if ocalfits_file is supplied so too must acalfits_file
-    if not(ocalfits_file is not None and acalfits_file is None) or (ocalfits_file is None and acalfits_file is not None):
+    if (ocalfits_file is not None and acalfits_file is None) or (ocalfits_file is None and acalfits_file is not None):
         raise ValueError("Both an abscal and omnical file must be supplied or none at all.")
+    if history is None:
+        raise ValueError("Must provide a non-empty history string.")
 
     history = 'Flagging command: "' + history + '", Using ' + hera_qm_version_str
     dirname = resolve_xrfi_path(xrfi_path, data_file, jd_subdir=True)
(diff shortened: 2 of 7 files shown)


hera_qm/xrfi.py
```diff
@@ -1373,9 +1373,10 @@ def chi_sq_pipe(uv, alg='zscore_full_array', modified=False, sig_init=6.0,
 #############################################################################
 
 
-def xrfi_run(ocalfits_file, acalfits_file, model_file, data_file, history,
+def xrfi_run(ocalfits_file=None, acalfits_file=None, model_file=None, data_file=None, history=None,
```
jsdillon, line 1376: For abscal and omnical, we need to be able to turn on/off gain-based flags and chi^2-based flags (both the med/meanfilt ones and the absolute ones using the chi_sq_pipe). For H3C/H4C, I think we'll probably want to turn off abscal chi^2 flagging, since it might get triggered when the RIMEz model fails.

hera_qm/xrfi.py
```diff
@@ -1442,164 +1448,239 @@ def xrfi_run(ocalfits_file, acalfits_file, model_file, data_file, history,
...
+        if len(flags) > 1:
+            for flg in flags[1:]:
+                uvf_init |= flg
+        uvf_init.label = 'ORd flags, round 1.'
+    else:
+        uvf_init = None
+        uvf_fws = None; uvf_f = None; uvf_metrics = None
 
+    if not data_file is None:
     # Second round -- use init flags to mask and recalculate everything
```
jsdillon, line 1578: missing indentation of comments

hera_qm/xrfi.py
```diff
@@ -1442,164 +1448,239 @@ def xrfi_run(ocalfits_file, acalfits_file, model_file, data_file, history,
...
-    uvf_ax2, uvf_axf2 = xrfi_pipe(uvc_a, alg='detrend_meanfilt', Kt=kt_size, Kf=kf_size, xants=xants,
-                                  cal_mode='tot_chisq', sig_init=sig_init, sig_adj=sig_adj,
-                                  label='Abscal chisq, round 2.',
-                                  run_check=run_check,
-                                  check_extra=check_extra,
-                                  run_check_acceptability=run_check_acceptability)
-
+    metrics = []
+    flags = []
+    if uvc_a is not None and uvc_o is not None:
```
jsdillon, line 1595: again, why are these tied?

hera_qm/xrfi.py
```diff
@@ -1983,34 +2068,40 @@ def day_threshold_run(data_files, history, nsig_f=7., nsig_t=7.,
...
             files1 = [glob.glob(d + '/*' + ext + '1.h5')[0] for d in xrfi_dirs]
             files2 = [glob.glob(d + '/*' + ext + '2.h5')[0] for d in xrfi_dirs]
             uvf1 = UVFlag(files1)
             uvf2 = UVFlag(files2)
             uvf2.metric_array = np.where(np.isinf(uvf2.metric_array), uvf1.metric_array,
                                          uvf2.metric_array)
             filled_metrics.append(uvf2)
-        else:
-            # Data was only run in second iteration
+        elif np.all([len(f) > 0 for f in files2_all]):
```
jsdillon, line 2083: Do we want to consider the possibility that some files will exist in round 1 but not round 2?

hera_qm/xrfi.py
```diff
@@ -1442,164 +1448,239 @@ def xrfi_run(ocalfits_file, acalfits_file, model_file, data_file, history,
     None
 
     """
+    if ocalfits_file is None and acalfits_file is None and model_file is None and data_file is None:
+        raise ValueError("Must provide at least one of the following; ocalfits_file, acalfits_file, model_file, data_file")
+    # if ocalfits_file is supplied so too must acalfits_file
+    if (ocalfits_file is not None and acalfits_file is None) or (ocalfits_file is None and acalfits_file is not None):
+        raise ValueError("Both an abscal and omnical file must be supplied or none at all.")
```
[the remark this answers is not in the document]
anguta: I agree if we can run the chi_sq pipe on whichever cal is provided and we can set a_priori flags to whichever cal is provided.

hera_qm/xrfi.py
```diff
@@ -1442,164 +1448,239 @@ def xrfi_run(ocalfits_file, acalfits_file, model_file, data_file, history,
...
-    uvf_ox, uvf_oxf = xrfi_pipe(uvc_o, alg='detrend_medfilt', Kt=kt_size, Kf=kf_size, xants=xants,
-                                cal_mode='tot_chisq', sig_init=sig_init, sig_adj=sig_adj,
-                                label='Omnical chisq, round 1.',
-                                run_check=run_check,
-                                check_extra=check_extra,
-                                run_check_acceptability=run_check_acceptability)
-
+    metrics = []
+    flags = []
+    if acalfits_file is not None and ocalfits_file is not None:
```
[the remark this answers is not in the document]
anguta: I think I was reading to much into the specific interdependency that existed in `xrfi_run` -- apriori flags requires abscal and chisq requires omnical in the current implementation. If we can go ahead and let these be calculated with whatever cal is provided then I agree.

hera_qm/xrfi.py
```diff
@@ -1442,164 +1448,239 @@ def xrfi_run(ocalfits_file, acalfits_file, model_file, data_file, history,
...
+        uvc_a = UVCal()
+        uvc_a.read_calfits(acalfits_file)
+        uvf_apriori = UVFlag(uvc_a, mode='flag', copy_flags=True, label='A priori flags.')
```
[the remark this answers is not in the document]
anguta: We can go with that. The current changes were meant to preserve the current structure of `xrfi_run` as much as possible -- probably too strictly.

hera_qm/xrfi.py
```diff
@@ -1373,7 +1380,15 @@ def chi_sq_pipe(uv, alg='zscore_full_array', modified=False, sig_init=6.0,
...
-def xrfi_run(ocalfits_file=None, acalfits_file=None, model_file=None, data_file=None, history=None,
+def xrfi_run(ocalfits_file=None, acalfits_file=None, model_file=None, data_file=None,
+             omnical_median_filter=True, abscal_median_filter=True,
+             omnical_chi2_median_filter=True, abscal_chi2_median_filter=True,
+             omnical_mean_filter=True, abscal_mean_filter=True,
+             omnical_chi2_mean_filter=True, abscal_chi2_mean_filter=True,
+             omnical_zscore_filter=True, abscal_zscore_filter=True,
+             omnivis_median_filter=True, data_median_filter=False,
+             omnivis_mean_filter=True, data_mean_filter=True,
+             correlations='both', history=None,
```
jsdillon, line 1391: I'm not a fan of this approach. I think we should consider the autos and crosses as fundamentally different data sources (certainly their statistics are quite different). The only reason weren't not doing a median filter on the data by default is speed. But we could easily do a median filter on the autos and then a mean filters on both autos and crosses (probably separately).

My suggestion would be to add `autos_median_filter=True` and `data_mean_filter=True` as defaults. We could also rename the `data` ones `crosses` or something like that.

hera_qm/xrfi.py
```diff
@@ -1398,10 +1413,70 @@ def xrfi_run(ocalfits_file=None, acalfits_file=None, model_file=None, data_file=
...
     acalfits_file : str, optional
         The abscal calfits file to use to flag on gains and chisquared values.
+    medfilt : bool, optional
+        Determine whether to run a round of median filtering
     model_file : str, optional
         THe model visibility file to flag on.
     data_file : str, optional
         The raw visibility data file to flag.
+    omnical_median_filter : bool, optional
+        If true, run a median filter on omnical gains.
```
jsdillon, line 1423: For all of these, I'd clarify that if the source file(s) is None, they aren't run even if True.

hera_qm/xrfi.py
```diff
@@ -1398,10 +1413,70 @@ def xrfi_run(ocalfits_file=None, acalfits_file=None, model_file=None, data_file=
         The omnical calfits file to use to flag on gains and chisquared values.
     acalfits_file : str, optional
         The abscal calfits file to use to flag on gains and chisquared values.
+    medfilt : bool, optional
```
jsdillon, line 1416: This option doesn't appear to be used.

hera_qm/xrfi.py
```diff
@@ -1469,79 +1544,166 @@ def xrfi_run(ocalfits_file=None, acalfits_file=None, model_file=None, data_file=
...
+                                        check_extra=check_extra,
+                                        run_check_acceptability=run_check_acceptability)
+            metrics += [uvf_ox]
+            flags += [uvf_oxf]
+        else:
+            uvf_ox = None
+            uvf_oxf = None
+        if omnical_zscore_filter:
+            uvf_oz, uvf_ozf = chi_sq_pipe(uvc_o, alg='zscore_full_array', modified=True,
+                                                   sig_init=sig_init, sig_adj=sig_adj,
```
jsdillon, line 1581: Tab issues

hera_qm/xrfi.py
```diff
@@ -1469,79 +1544,166 @@ def xrfi_run(ocalfits_file=None, acalfits_file=None, model_file=None, data_file=
...
-        # Get the absolute chi-squared values
-        uvf_chisq, uvf_chisq_f = chi_sq_pipe(uvc_o, alg='zscore_full_array', modified=True,
-                                             sig_init=sig_init, sig_adj=sig_adj,
-                                             label='Renormalized chisq, round 1.',
-                                             run_check=run_check,
-                                             check_extra=check_extra,
-                                             run_check_acceptability=run_check_acceptability)
-        metrics += [uvf_chisq]
-        flags += [uvf_chisq_f]
+        if omnical_median_filter:
```
jsdillon, line 1555: I like this new organization a lot, but I do wonder whether this would be easier to read and maintain if you make a few helper functions to modularize this whole function better. Right now it's very long and there's a lot of copy-pasted code.

I'm also OK if you make this an issue and defer it to a future PR if that makes more sense to you.

hera_qm/xrfi.py
```diff
@@ -1574,94 +1736,120 @@ def xrfi_run(ocalfits_file=None, acalfits_file=None, model_file=None, data_file=
...
+                                          run_check=run_check,
+                                          check_extra=check_extra,
+                                          run_check_acceptability=run_check_acceptability)
+            metrics += [uvf_ox2]
+            flags += [uvf_oxf2]
+        else:
+            uvf_ox2 = None; uvf_oxf2 = None
+        if omnical_zscore_filter:
+            uvf_oz2, uvf_ozf2 = chi_sq_pipe(uvc_o, alg='zscore_full_array', modified=True,
+                                                   sig_init=sig_init, sig_adj=sig_adj,
```
jsdillon, line 1814: More undertabbing

hera_qm/xrfi.py
anguta: I've untied them.

hera_qm/xrfi.py
anguta: I now consider this possibility as well.

hera_qm/xrfi.py
anguta: I agree. We should be able to median filter the autos but only mean filter the crosses. I've changed the structure to allow this, putting autos/crosses on the same level as omnivis, abscal, omnical.

hera_qm/xrfi.py
anguta: I've added this clarification.

hera_qm/xrfi.py
anguta: I agree that this could probably be more modularized. I think we should defer to a different PR. I've made an issue #346

hera_qm/xrfi.py
anguta: i've gotten rid of this.

hera_qm/xrfi.py
anguta: thanks for catching this. It's fixed.

hera_qm/xrfi.py
anguta: fixed.

hera_qm/xrfi.py
anguta: Fixed.

hera_qm/xrfi.py
```diff
@@ -1442,164 +1448,239 @@ def xrfi_run(ocalfits_file, acalfits_file, model_file, data_file, history,
     None
 
     """
+    if ocalfits_file is None and acalfits_file is None and model_file is None and data_file is None:
+        raise ValueError("Must provide at least one of the following; ocalfits_file, acalfits_file, model_file, data_file")
+    # if ocalfits_file is supplied so too must acalfits_file
+    if (ocalfits_file is not None and acalfits_file is None) or (ocalfits_file is None and acalfits_file is not None):
+        raise ValueError("Both an abscal and omnical file must be supplied or none at all.")
+    if history is None:
+        raise ValueError("Must provide a non-empty history string.")
```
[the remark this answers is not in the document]
anguta: I now use '' as default.

hera_qm/xrfi.py
```diff
@@ -1426,6 +1430,8 @@ def xrfi_run(ocalfits_file, acalfits_file, model_file, data_file, history,
         Metrics file that contains a list of excluded antennas. Flags of visibilities
         formed with these antennas will be set to True. Default is None (i.e.,
         no antennas will be excluded).
+    output_prefix : str, optional
+        Optional output prefix. If none is provided, use data-file.
```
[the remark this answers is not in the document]
anguta: Added a bit of clarifying documetation.

hera_qm/utils.py
```diff
@@ -276,6 +276,36 @@ def get_metrics_ArgumentParser(method_name):
...
+                        'algorithm. Default is 8.')
+        ap.add_argument('--kf_size', default=8, type=int,
+                        help='Size of kernel in frequency dimension for detrend in '
+                        'xrfi algorithm. Default is 8.')
+        ap.add_argument('--sig_init', default=6.0, type=float,
+                        help='Starting number of sigmas to flag on. Default is 6.0.')
+        ap.add_argument('--sig_adj', default=2.0, type=float,
+                        help='Number of sigmas to flag on for data adjacent to a flag. Default is 2.0.')
+        ap.add_argument('--ex_ants', default=None, type=str,
+                        help='Comma-separated list of antennas to exclude. Flags of visibilities '
```
jsdillon, line 296: This only works if there are no spaces right? Should probably clarify that.

hera_qm/utils.py
```diff
@@ -326,6 +356,8 @@ def get_metrics_ArgumentParser(method_name):
         ap.add_argument("--run_if_first", default=None, type=str, help='only run \
                         day_threshold_run if the first item in the sorted data_files \
                         list matches run_if_first (default None means always run)')
+        ap.add_argument("--skip_making_flagged_abs_calfits", default=False, action="store_true",
+                        help='If True, flag the abscal files.')
```
jsdillon, line 360: This help string has things backwards, I think

scripts/xrfi_run_data_only.py
```diff
@@ -0,0 +1,20 @@
...
+ap = utils.get_metrics_ArgumentParser('xrfi_run_data_only')
+args = ap.parse_args()
+history = ' '.join(sys.argv)
+
+xrfi.xrfi_run(data_file=args.data_file,
+              cross_median_filter=args.cross_median_filter,
+              cross_mean_filter=not(args.skip_cross_mean_filter),
+              history=history, xrfi_path=args.xrfi_path,
+              kt_size=args.kt_size, kf_size=args.kf_size, sig_init=args.sig_init,
+              sig_adj=args.sig_adj, ex_ants=args.ex_ants, ant_str=args.ant_str,
```
jsdillon, line 19: ant_str was removed as an argument, I think

hera_qm/xrfi.py
```diff
@@ -1090,6 +1090,11 @@ def calculate_metric(uv, algorithm, cal_mode='gain', correlations='both', run_ch
         is raised.
 
     """
+    if issubclass(uv.__class__, (UVData)):
+        if correlations == 'auto':
+            uv = uv.select(bls=[(ant1, ant2) for ant1, ant2, in zip(uv.ant_1_array, uv.ant_2_array) if ant1 == ant2], inplace=False)
+        if correlations == 'cross':
+            uv = uv.select(bls=[(ant1, ant2) for ant1, ant2, in zip(uv.ant_1_array, uv.ant_2_array) if ant1 != ant2], inplace=False)
```
jsdillon, line 1097: I think you can do this select with `ant_str` instead of `bls` to streamline this code a bit.

hera_qm/xrfi.py
```diff
@@ -1381,16 +1389,18 @@ def chi_sq_pipe(uv, alg='zscore_full_array', modified=False, sig_init=6.0,
...
+             omnivis_median_filter=True, omnivis_mean_filter=True,
+             auto_median_filter=True, auto_mean_filter=True,
+             cross_median_filter=False, cross_mean_filter=True,
+             history=None,
              xrfi_path='', kt_size=8, kf_size=8, sig_init=5.0, sig_adj=2.0,
-             ex_ants=None, ant_str=None, metrics_file=None,
+             ex_ants=None, metrics_file=None,
              output_prefix=None, clobber=False,
              run_check=True, check_extra=True, run_check_acceptability=True):
     """Run the xrfi excision pipeline used for H1C IDR2.2.
```
jsdillon, line 1406: This line is no longer true. Maybe "originally developed for H1C IDR2.2" or something like that

hera_qm/utils.py
anguta: Thanks for catching this!

hera_qm/xrfi.py
anguta: good point. Done this.

scripts/xrfi_run_data_only.py
anguta: i've removed ant_str here and in the parsers.

hera_qm/xrfi.py
anguta: which part of this line are you referring too?

hera_qm/utils.py
anguta: Do you mean that `ex_ants` won't work if there are not spaces in the comma-separated list of antennas? I believe the other ex_ants parsers also neglect to mention this. I can update all of them if you want .

jsdillon: Nevermind, it works regardless of spaces as long as there are also commas.

hera_qm/xrfi.py
jsdillon: Weird, I meant to reference the line `Run the xrfi excision pipeline used for H1C IDR2.2.` but it seems to have gotten misattributed.

jsdillon: I'll just fix it myself.

PR #348, opened 2020-07-26
Add missing scripts to setup

jsdillon:
New XRFI run scripts were not getting installed with the module.


jsdillon:
Something has gone awry with the unit tests. Starting over...

Issue #351, opened 2020-07-28
XRFI no longer accepts multiple files 


jsdillon:
https://github.com/HERA-Team/hera_qm/blob/5313e1f41bc256ad5e9ca163039515f5c24956ae/hera_qm/utils.py#L278

This is needed for a chunked analysis.

jsdillon:
Closed by #352

PR #352, opened 2020-07-29
Xrfi multi file

anguta:
addresses #351 and adds an explit multifile unittest.


hera_qm/utils.py
```diff
@@ -245,14 +245,18 @@ def get_metrics_ArgumentParser(method_name):
...
+                        nargs='+')
+        ap.add_argument('--acalfits_files', default=None, type=str, help='Abscal '
+                        'calfits files to use to flag on gains and chisquared values.',
+                        nargs='+')
+        ap.add_argument('--model_files', default=None, type=str, help='Model visibility '
+                        'file to flag on.',
+                        nargs='+')
         ap.add_argument('--data_file', default=None, type=str, help='Raw visibility '
-                        'data file to flag on.')
+                        'data files to flag on.',
```
jsdillon, line 258: If the other variables are being pluralized, this one should be too.

hera_qm/xrfi.py
```diff
@@ -1975,13 +1984,50 @@ def xrfi_run(ocalfits_file=None, acalfits_file=None, model_file=None, data_file=
...
-            uvf.write(outpath, clobber=clobber)
 
+    # Determine the actual files to store
+    # We will drop kt_size / (integrations per file) files at the start and
+    # end to avoid edge effects from the convolution kernel.
+    # If this chunk includes the start or end of the night, we will write
+    # output files for those, but flag everything.
+
+    # Read metadata from first file to get integrations per file.
+    file_lists = [ocalfits_files, acalfits_files, data_files, model_files]
```
jsdillon, line 1995: Not sure what this line accomplishes.

hera_qm/xrfi.py
```diff
@@ -1975,13 +1984,50 @@ def xrfi_run(ocalfits_file=None, acalfits_file=None, model_file=None, data_file=
...
+    if data_files is not None:
```
jsdillon, line 1996: This series of conditionals seems a bit convoluted. Not sure how to make it much better, but maybe we can dispense with dtypt and just have a uvtemp = UVData/Cal and uvtemp.read() inside the first set of conditionals. I think that would be cleaner and a bit easier to follow.

hera_qm/xrfi.py
```diff
@@ -1975,13 +1984,50 @@ def xrfi_run(ocalfits_file=None, acalfits_file=None, model_file=None, data_file=
...
+    for ind in range(0, len(uvlist)):
+        dirname = resolve_xrfi_path(xrfi_path, output_prefixes[ind], jd_subdir=True)
+        basename = qm_utils.strip_extension(os.path.basename(output_prefixes[ind]))
+        for ext, uvf in uvf_dict.items():
+            if uvf is not None:
+                # This is calculated separately for each uvf because machine
+                # precision error was leading to times not found in object.
+                this_times = np.unique(uvf.time_array)
+                t_ind = ind * uvtemp.Ntimes
+                uvf_out = uvf.select(times=this_times[t_ind:(t_ind + uvtemp.Ntimes)],
```
jsdillon, line 2025: This assumes that all files have the same number of integrations. Do we think that might be a problem? Maybe this is a question for @plaplant?

I'd at least document the assumption and maybe raise an issue if we think it'll be a problem in the near term...

hera_qm/xrfi.py
```diff
@@ -1975,13 +1984,50 @@ def xrfi_run(ocalfits_file=None, acalfits_file=None, model_file=None, data_file=
...
+        dtypet = 'uvcal'
+        uvlist = acalfits_files
+    if dtypet == 'uvdata':
+        uvtemp = UVData()
+        uvtemp.read(uvlist[0], read_data=False)
+    elif dtypet == 'uvcal':
+        uvtemp = UVCal()
+        uvtemp.read_calfits(uvlist[0])
+
+    # Loop through the files to output, storing all the different data products.
```
jsdillon, line 2015: It looks like your dropped the feature where you only output files for central files. @adampbeardsley's idea here was to try to avoid edge-effects by deriving flags for a set of files using a slightly larger set of files and then not writing down the flags from the results from the edge files. The new OPM behavior lets you do this kind of chunking where the chunk size is larger than the stride length, allowing for flagging on overlapping sets of files. 

I think this behavior should be on by default when you have multiple files, but I'm open to it being default false. Regardless, we need it to be triggerable from the command line with a boolean flag. 

See https://github.com/HERA-Team/hera_qm/blob/5313e1f41bc256ad5e9ca163039515f5c24956ae/hera_qm/xrfi.py#L2226-L2266
for the implementation

hera_qm/tests/test_xrfi.py
```diff
@@ -1174,6 +1183,101 @@ def test_xrfi_run(tmpdir):
...
+    with pytest.warns(None) as record:
+        xrfi.xrfi_run(ocal_files, acal_files, model_files, raw_dfiles,
+                      'Just a test', kt_size=3, cross_median_filter=True)
+    assert len(record) >= len(messages)
+    n_matched_warnings = 0
+    for i in range(len(record)):
+        if mess1[0] in str(record[i].message) and cat1[0] == record[i].category:
+            n_matched_warnings += 1
+    assert n_matched_warnings == 8
+    outdirs = ['zen.2457698.40355191.xrfi',
```
jsdillon, line 1225: If you add back in the feature for not flagging edge files, then I think we should run this twice, once with it on and once with it off. In the first case, this should produce one output file (depending on Kt) and in the later case, this should produce 3 files.

hera_qm/utils.py
anguta: yep.

hera_qm/xrfi.py
anguta: orphaned. Will remove.

hera_qm/xrfi.py
anguta: I'll document it for now.

hera_qm/xrfi.py
anguta: I agree.

hera_qm/xrfi.py
anguta: I ended up leaving this out because it broke the day_thresholding unittest and it wasn't clear to me why it existed. Now that you've explained it, I see it's use. I can add it back in and disable it on day thresholding.

hera_qm/tests/test_xrfi.py
anguta: Good idea. I'll add this test.

hera_qm/xrfi.py
jsdillon: Thanks! Happy to review whenever you're ready.

anguta: should be ready now.

hera_qm/tests/test_xrfi.py
anguta: i end up checking whether all times in flag zone are flagged since there aren't enough integrations to test whether files haven't been written.

anguta:
Thanks for the review @jsdillon. Should be ready for another look.

hera_qm/xrfi.py
```diff
@@ -1540,6 +1540,12 @@ def xrfi_run(ocalfits_files=None, acalfits_files=None, model_files=None, data_fi
         (should have a .uvh5 at the end). Output products will replace extension
         with various output labels. For example, output_prefixes='filename.uvh5'
         will result in products with names like 'filename.cross_flags1.h5'.
+    throw_away_edges : bool, optional
```
jsdillon, line 1543: My only suggestion was to expand this doc string, so I just did it. If you're OK with it, then feel free to merge.

PR #358, opened 2020-08-07
apriori frequency flags

### LINKED ISSUE #332: Enable XRFI to take a priori frequency flags
After some discussion in the most recent analysis telecon, it was suggested that we should provide a simple interface (perhaps a text file) specifying ranges in frequency to always flag. 

These should be integrated into XRFI in a self-consistent way, including treating them as high nsigma points so that the watershed algorithm can grow flags around them.

anguta:
Closes #332 

apriori flags are provided as a simple comma separated text file with '#' denoting comment lines and each row has the format

lower_bound, upper_bound

### DIFF
--- a/hera_qm/utils.py
+++ b/hera_qm/utils.py
@@ -633,7 +633,7 @@ def read_bounds_text_file(list_file=None):
                         lower = -np.inf
                     else:
                         lower = float(lower)
-                    if upper == '':
+                    if upper == '' or upper == '\n':
                         # if no upper is provided, set to infinity.
                         upper = np.inf
                     else:
--- a/hera_qm/utils.py
+++ b/hera_qm/utils.py
@@ -601,3 +601,44 @@ def strip_extension(path, return_ext=False):
         return (root, ext[1:])
     else:
         return os.path.splitext(path)[0]
+
+def read_bounds_text_file(list_file=None):
+    """Read in apriori flags
+
+    This method reads in a plaintext comma-separated file
+    that lists lower and upper bounds on time or frequency regions
+    to flag apriori. Returns a list of 2-tuples with these lower / uppper bounds.
+
+    Parameters
+    ----------
+    list_file : str, optional
+        Path to list_file to parse.
+        Default, None. Returns empty list.
+
+    Returns
+    -------
+    region_list : list of 2-tuples indicating upper and lower bounds of regions retrieved from text file.
+    """
+    if list_file is not None:
+        with open(list_file, 'r') as input:
+            regions = []
+            lines = input.readlines()
+            for line in lines:
+                line = line.replace(' ', '')
+                line = line.replace('\t', '')
+                if line[0] != '#':
+                    lower, upper = line.split(',')
+                    if lower == '':
+                        # if no lower bound is provided, set to negative infinity.
+                        lower = -np.inf
+                    else:
+                        lower = float(lower)
+                    if upper == '':
+                        # if no upper is provided, set to infinity.
+                        upper = np.inf
+                    else:
+                        upper = float(upper)
+                    regions.append((min(lower, upper), max(lower, upper)))
+        return regions
+    else:
+        return []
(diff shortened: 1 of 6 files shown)


hera_qm/utils.py
```diff
@@ -601,3 +601,44 @@ def strip_extension(path, return_ext=False):
         return (root, ext[1:])
     else:
         return os.path.splitext(path)[0]
+
+def read_bounds_text_file(list_file=None):
+    """Read in apriori flags
+
+    This method reads in a plaintext comma-separated file
```
jsdillon, line 608: Calling it "comma-separated" is confusing, since it uses both commas and new-lines as delimiters. I'd be more explicit here.

It also occurs to me what we might want to use tomls (which support notions of nested arrays). I'm worried we're reinventing the wheel here.

hera_qm/utils.py
```diff
@@ -601,3 +601,44 @@ def strip_extension(path, return_ext=False):
...
+    """
+    if list_file is not None:
+        with open(list_file, 'r') as input:
+            regions = []
+            lines = input.readlines()
+            for line in lines:
+                line = line.replace(' ', '')
+                line = line.replace('\t', '')
+                if line[0] != '#':
+                    lower, upper = line.split(',')
```
jsdillon, line 630: I'd put in some error checking to make sure that this that each line produces exactly 2 entries.

hera_qm/utils.py
```diff
@@ -601,3 +601,44 @@ def strip_extension(path, return_ext=False):
...
+                    if lower == '':
+                        # if no lower bound is provided, set to negative infinity.
+                        lower = -np.inf
```
jsdillon, line 633: Do we really need to support a format like this? I'm fine with forcing people putting in 0 for one side and whatever large number they want for the other...

hera_qm/utils.py
```diff
@@ -601,3 +601,44 @@ def strip_extension(path, return_ext=False):
...
     else:
         return os.path.splitext(path)[0]
+
+def read_bounds_text_file(list_file=None):
+    """Read in apriori flags
+
+    This method reads in a plaintext comma-separated file
+    that lists lower and upper bounds on time or frequency regions
+    to flag apriori. Returns a list of 2-tuples with these lower / uppper bounds.
+
```
jsdillon, line 611: Clarify that they are in the same units as the data files, so usually Hz.

hera_qm/utils.py
jsdillon: If we use a toml, this could be easily extended to channel numbers (and maybe even integration numbers / JDs / LSTs). Since a toml lets us define variable arrays, we could define one array as freq flags (using nested arrays). We could define another one for channel flags which could be a mixed nested array:

`channel_flags = [0, 1, [10, 20], 60]`

which would flag channels 0, 1, 10, 11, ... 19, 20, 60. 

We could even do both, freq_flags and channel_flags, with the caveat that the output of read_bounds will always be either channels or flags (your choice) and will require a frequency array to convert them if other type (or both) is provided.

This could definitely be extended to times in a way that supports unified I/O and it feels a lot more "pythonic" than a slightly janky text file. What do you think?

jsdillon:
To illustrate what I mean about tomls, I mocked up a format that we might use. I don't think we need to implement the time flagging stuff in this PR, but I just wanted to show it we might extend this in the toml format:

```
# A Priori Flag File

[frequency_flags]
# freq_flags are an array of length-2 arrays of floats in units of Hz
freq_flags = [[0, 110e6], [150e6, 155e6], [190e6, 200e6]]
# channel_flags are an array of ints or length-2 arrays of ints
channel_flags = [0, 1, [10, 20], 60]

[time_flags]
# JD_flags are an array of length-2 arrays of floats in units of days.
JD_flags = [[2458838.01, 2458838.03]]
# LST_flags are an array of length-2 arrays of floats in units of hours. If LST_1 > LST_2, it's assumed to span the 24 branch cut.
LST_flags = [[2.3, 3.5], [16.2, 18.4], [23.4, 0.1]]
# integration_flags are an array of ints or length-2 arrays of ints
integration_flags = [0, 1, [10, 20], 60]
```

PR #360, opened 2020-08-10
Apply a priori flag yaml

anguta:
Takes the yaml machinery added in #359 and applies to to `xrfi_run` to pre-flag frequencies and times. This pull request closes #358 .


hera_qm/data/a_priori_flags_integrations.yaml
```diff
@@ -0,0 +1,40 @@
...
+# channel_flags are an array of ints or length-2 arrays of ints (channel ranges, inclusive)
+channel_flags: [0, 1, [10, 20], 60]
+
+
+# ---------------
+# TIME FLAGS
+# ---------------
+
+# # JD_flags are an array of length-2 arrays of floats in units of days (time ranges, inclusive)
+JD_flags: []
```
jsdillon, line 25: FYI, instead of leaving them empty you can also choose not to have these fields. That's your call.

hera_qm/utils.py
```diff
@@ -610,3 +613,66 @@ def strip_extension(path, return_ext=False):
         return (root, ext[1:])
     else:
         return os.path.splitext(path)[0]
+
+
+def apply_yaml_freq_time_flags(uv, a_priori_flag_yaml, lat_lon_alt_degrees=None, telescope_name='HERA'):
```
jsdillon, line 618: You may want to get the telescope name from the `uv` object, but then allow the user to overwrite it (so default None).

hera_qm/utils.py
```diff
@@ -610,3 +613,66 @@ def strip_extension(path, return_ext=False):
...
+
+        # calculate LST grid in hours from time grid and lat_lon_alt
+        lst_array = uvutils.get_lst_for_time(uv.time_array, *lat_lon_alt_degrees) * 12 / np.pi
+    elif issubclass(uv.__class__, UVData):
+        lst_array = np.unique(uv.lst_array) * 12  / np.pi
+    else:
+        raise NotImplementedError("uv must be a UVData or UVCal object.")
+    time_array = np.unique(uv.time_array)
+    # loop over spws to apply frequency flags.
+    for spw in range(uv.Nspws):
```
jsdillon, line 659: While this will work for freq_flags just fine, I think this will break for channel flags. I'd suggest just raising a NotImplementedError if Nspws > 1 to sidestep this issue.

hera_qm/utils.py
```diff
@@ -610,3 +613,66 @@ def strip_extension(path, return_ext=False):
...
+        flagged_channels = metrics_io.read_a_priori_chan_flags(a_priori_flag_yaml, freqs=uv.freq_array[spw])
+        flagged_channels = flagged_channels[(flagged_channels>=0) & (flagged_channels<=uv.Nfreqs)]
```
jsdillon, line 661: Should we put this check into `read_a_priori_chan_flags`? Perhaps with a warning if flags are being dropped?

jsdillon: Certainly we can put the channel flags >= 0 pretty easily by having it error if they are < 0

hera_qm/utils.py
```diff
@@ -610,3 +613,66 @@ def strip_extension(path, return_ext=False):
...
+        if issubclass(uv.__class__, UVData):
+            uv.flag_array[:, spw, flagged_channels, :] = True
+        elif issubclass(uv.__class__, UVCal):
+            uv.flag_array[:, spw, flagged_channels, :, :] = True
+    # now do times.
+    # get the integrations to flag
+    flagged_integrations = metrics_io.read_a_priori_int_flags(a_priori_flag_yaml, lsts=lst_array, times=time_array)
+    # ony select integrations less then Ntimes
+    flagged_integrations = flagged_integrations[(flagged_integrations>=0) & (flagged_integrations<=uv.Ntimes)]
```
jsdillon, line 670: Likewise this one, can we fold some of this check into read_a_priori_int_flags?

hera_qm/xrfi.py
```diff
@@ -1490,6 +1491,10 @@ def xrfi_run_step(uv_files=None, uv=None, uvf_apriori=None,
         if True, calculate modified_z_score when computing overall z-score
         (only used if alg='overall_z_score').
         Default is False.
+    a_priori_flag_yaml : str, optional
```
jsdillon, line 1494: I'm having a hard time telling, but do the flags that are in the YAML get propagated to `uvf_apriori`? If not, they definitely should since that gets written to disk (which will provide a really nice check that things are working as expected).

hera_qm/utils.py
```diff
@@ -610,3 +613,66 @@ def strip_extension(path, return_ext=False):
...
+
+def apply_yaml_freq_time_flags(uv, a_priori_flag_yaml, lat_lon_alt_degrees=None, telescope_name='HERA'):
+    """Apply frequency and time flags to a UVData or UVCal object
+
+    This function takes in a uvdata or uvcal object and applies
+    frequency and time flags as appropriate.
+
+    Parameters
+    ----------
+    uv : UVData or UVCal object
```
jsdillon, line 626: (or subclass thereof)

hera_qm/utils.py
jsdillon, line 618: Does it also make sense to apply antenna flags using this same function? I know it'd be redundant in xrfi, but I think it would be useful for internal logical consistency and future usefulness of this function.

hera_qm/utils.py
```diff
@@ -610,3 +613,66 @@ def strip_extension(path, return_ext=False):
...
+        uvdata or uvcal to apply frequency / time flags to.
+    a_priori_flag_yaml : str
+        path to yaml file with frequeny / time flags.
+    lat_lon_alt_degrees : list, optional
+        list of latitude, longitude, and altitude for telescope.
```
jsdillon, line 631: I'd clarify that alt is in meters.

hera_qm/xrfi.py
anguta: The flags do get propagated to `uvf_apriori` this is something the unittests check. They do not necessarily get propagated to metric derived flags since the metrics are computed in flagged channels?

hera_qm/utils.py
anguta: Added this.

hera_qm/utils.py
anguta: I agree. We might as well add that here.

hera_qm/utils.py
anguta: added clarification.

hera_qm/data/a_priori_flags_integrations.yaml
anguta: I'll just leave them empty.

hera_qm/utils.py
```diff
@@ -610,3 +613,66 @@ def strip_extension(path, return_ext=False):
...
+        Default is HERA
+
+    Returns
+    -------
+        uv : UVData or UVCal object
+            input uvdata / uvcal but now with flags applied
+    """
+    # if UVCal provided, get lst_array from times.
+    # If lat_lon_alt is not specified, try to infer it from the telescope name, which calfits files generally carry around
+    if issubclass(uv.__class__, UVCal):
```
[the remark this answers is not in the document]
anguta: changed this.

hera_qm/utils.py
```diff
@@ -610,3 +613,66 @@ def strip_extension(path, return_ext=False):
...
+        if lat_lon_alt_degrees is None:
+            if telescope_name.upper() == 'HERA':
+                lat_lon_alt_degrees = np.array(uvutils.LatLonAlt_from_XYZ(HERA_TELESCOPE_LOCATION))
+                lat_lon_alt_degrees *= [180 / np.pi, 180 / np.pi, 1]
```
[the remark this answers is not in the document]
anguta: I've adopted this.

hera_qm/utils.py
anguta: changed default to None. If none, obtained from `uv`.

hera_qm/utils.py
```diff
@@ -19,7 +20,9 @@ def _bytes_to_str(inbyte):
 def _str_to_bytes(instr):
     return instr.encode('utf8')
 
-
+HERA_TELESCOPE_LOCATION = np.array([5109325.855210627429187297821044921875,
```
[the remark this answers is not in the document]
anguta: I've removed this.

hera_qm/utils.py
anguta: The behavior here will be to apply the same channel flags for different spws. If this behavior is weird, then lets drop it.

hera_qm/data/a_priori_flags_integrations.yaml
anguta: leaving it empty.

hera_qm/utils.py
anguta: I don't think this check can always be performed in `read_a_priori_chan_flags` since frequencies are not necessarily provided so it will be tricky too determine if the channel flags exceed maximum channels. I will add a warning here though that flags are being dropped in this method though.

hera_qm/utils.py
anguta: yep.

hera_qm/utils.py
jsdillon: That is weird... but if you think it's better than an error, I'd be OK with a warning.

hera_qm/utils.py
```diff
@@ -610,3 +611,118 @@ def strip_extension(path, return_ext=False):
...
+    for spw in range(uv.Nspws):
+        flagged_channels = metrics_io.read_a_priori_chan_flags(a_priori_flag_yaml, freqs=uv.freq_array[spw])
+        if np.any(flagged_channels >= uv.Nfreqs):
+            warnings.warn("Flagged channels were provided that exceed the maximum channel index. These flags are being dropped!")
+        if np.any(flagged_channels < 0):
+            warnings.warn("Flagged channels were provided with a negative channel index. These flags are being dropped!")
+        flagged_channels = flagged_channels[(flagged_channels>=0) & (flagged_channels<=uv.Nfreqs)]
+        if issubclass(uv.__class__, UVData):
+            uv.flag_array[:, spw, flagged_channels, :] = True
+        elif issubclass(uv.__class__, UVCal):
```
jsdillon, line 684: Since we're no longer enforcing UVData or UVCal above, we should raise an error if nothing is happening here.  

Also, do we want to support UVFlag too? It's a bit annoying, since UVFlag has 3 different modes, but one of them looks like UVData, one looks like UVCal, and one is just a waterfall.

hera_qm/xrfi.py
jsdillon: If these flags are applied to the UVCal/UVData object before calculating the metric, then they should get propagated to the other flag objects. Not 100% sure about that, but I don't think we need to worry about that now. As long as it appears in uvf_apriori, I'm happy.

anguta: Sounds good.

hera_qm/utils.py
anguta: I did think it might be nice to add `UVFlag` but I think let's make it an issue for a different PR if the need arises.

hera_qm/utils.py
anguta: Yeah it is weird. I'm going to make it throw an error :P. In the future, if we want too, we can add spws to the yaml format for channels to really include the support.

PR #371, opened 2020-09-03
Allow flagging yamls with no integrations, channels, and/or ants.

anguta:
Fix bug from slicing different lengthed lists of indices on multiple dimensions.

### DIFF
--- a/hera_qm/utils.py
+++ b/hera_qm/utils.py
@@ -725,11 +725,11 @@ def apply_yaml_flags(uv, a_priori_flag_yaml, lat_lon_alt_degrees=None, telescope
         if issubclass(uv.__class__, UVData):
             blt_selection = np.logical_or(uv.ant_1_array == antnum, uv.ant_2_array == antnum)
             if np.count_nonzero(blt_selection) > 0:
-                uv.flag_array[blt_selection, :, :, pol_selection] = True
+                uv.flag_array[blt_selection][:, :, :, pol_selection] = True
         elif issubclass(uv.__class__, UVCal):
             ant_selection = uv.ant_array == antnum
             if np.count_nonzero(ant_selection) > 0:
-                uv.flag_array[ant_selection, :, :, :, pol_selection] = True
+                uv.flag_array[ant_selection][:, :, :, :, pol_selection] = True
 
     # return uv with flags applied.
     return uv
--- a/hera_qm/utils.py
+++ b/hera_qm/utils.py
@@ -682,10 +682,11 @@ def apply_yaml_flags(uv, a_priori_flag_yaml, lat_lon_alt_degrees=None, telescope
         if np.any(flagged_channels < 0):
             warnings.warn("Flagged channels were provided with a negative channel index. These flags are being dropped!")
         flagged_channels = flagged_channels[(flagged_channels>=0) & (flagged_channels<=uv.Nfreqs)]
-        if issubclass(uv.__class__, UVData):
-            uv.flag_array[:, spw, flagged_channels, :] = True
-        elif issubclass(uv.__class__, UVCal):
-            uv.flag_array[:, spw, flagged_channels, :, :] = True
+        if len(flagged_channels) > 0:
+            if issubclass(uv.__class__, UVData):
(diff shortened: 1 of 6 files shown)


hera_qm/xrfi.py
```diff
@@ -1648,7 +1648,7 @@ def xrfi_run_step(uv_files=None, uv=None, uvf_apriori=None,
             # if uvf_apriori is not None and we wish to apply it to uv, do so here.
             elif apply_uvf_apriori:
                 flag_apply(uvf_apriori, uv, keep_existing=True, run_check=run_check,
-                           check_extra=check_extra,
+                           check_extra=check_extra, force_pol=True,
```
jsdillon, line 1651: I'm wondering whether there's ever a reason why we wouldn't want to `force_pol=True`. I can't think of any, but if you can, maybe expose this to the larger function with `True` as the default?

hera_qm/utils.py
```diff
@@ -722,10 +724,13 @@ def apply_yaml_flags(uv, a_priori_flag_yaml, lat_lon_alt_degrees=None, telescope
...
-            uv.flag_array[blt_selection, :, :, pol_selection] = True
+            if np.count_nonzero(blt_selection) > 0:
+                for bltind in np.where(blt_selection)[0]:
+                    uv.flag_array[bltind, :, :, pol_selection] = True
         elif issubclass(uv.__class__, UVCal):
             ant_selection = uv.ant_array == antnum
-            uv.flag_array[ant_selection, :, :, :, pol_selection] = True
-
+            if np.count_nonzero(ant_selection) > 0:
+                for antind in np.where(ant_selection)[0]:
```
jsdillon, line 733: Is there a reason you can't do `uv.flag_array[ant_selection, :, :, :, pol_selection] = True`?

hera_qm/utils.py
```diff
@@ -722,10 +724,13 @@ def apply_yaml_flags(uv, a_priori_flag_yaml, lat_lon_alt_degrees=None, telescope
             antnum = ant[0]
         if issubclass(uv.__class__, UVData):
             blt_selection = np.logical_or(uv.ant_1_array == antnum, uv.ant_2_array == antnum)
-            uv.flag_array[blt_selection, :, :, pol_selection] = True
+            if np.count_nonzero(blt_selection) > 0:
```
jsdillon, line 727: I prefer `np.any(blt_selection)` but that's stylistic.

hera_qm/utils.py
```diff
@@ -722,10 +724,13 @@ def apply_yaml_flags(uv, a_priori_flag_yaml, lat_lon_alt_degrees=None, telescope
...
+                for bltind in np.where(blt_selection)[0]:
+                    uv.flag_array[bltind, :, :, pol_selection] = True
         elif issubclass(uv.__class__, UVCal):
             ant_selection = uv.ant_array == antnum
-            uv.flag_array[ant_selection, :, :, :, pol_selection] = True
-
+            if np.count_nonzero(ant_selection) > 0:
```
jsdillon, line 732: I prefer `np.any(blt_selection)` but that's stylistic.

jsdillon (commented):
Basically looks good. Just a few questions @anguta

hera_qm/xrfi.py
anguta: `force_pol=True` is invoked here because `uvf_apriori` is a waterfall and will cause errors if `uv` has more then one pol. This made it past the unittests for some reason (are the unittests only involving single pol files?)

hera_qm/utils.py
anguta: I changed to this approach because `numpy` was throwing errors when the slice lengths don't match. This may be a version specific problem. With this change it is no longer an issue.

hera_qm/utils.py
anguta: I think your way is a bit more concise so I'm happy to change this.

hera_qm/utils.py
anguta: I agree.

jsdillon (approved):

PR #372, opened 2020-09-05
xrfi_run with no z-scores

anguta:


jsdillon:
Sorry to butt in here, but this seems like it should be an option with xrfi_run.py and not an entirely new script. We've got an over-proliferation of scripts here and I actually think we should be eliminating some of the older, unused ones (also some of the legacy XRFI functions).

jsdillon (approved):
Looks like you get mostly done with this @anguta, so I finished it. LMK if you're happy with my changes and we can merge in.

anguta (commented):
Looks good!

PR #374, opened 2020-09-10
Fix docstring

### LINKED ISSUE #332: Enable XRFI to take a priori frequency flags

### LINKED ISSUE #373: Add flags to xrfi_run.py to turn off various input sources
We need flags like `--ignore_omnical_median_filter` to turn off each of the default on sources of flags.

We probably also want `--use_cross_median_filter`to turn on the one default off source.

Related: this docstring is wrong:
https://github.com/HERA-Team/hera_qm/blob/18e1741c1b91ec9c7aee2edfaf8fc484ba497320/hera_qm/xrfi.py#L1810

jsdillon:
This closes #373

### DIFF
--- a/hera_qm/xrfi.py
+++ b/hera_qm/xrfi.py
@@ -1807,7 +1807,7 @@ def xrfi_run(ocalfits_files=None, acalfits_files=None, model_files=None,
     cross_median_filter : bool, optional
         If True, flag on data median filter statistic.
         Mean filters are run after median filters.
-        Default is True.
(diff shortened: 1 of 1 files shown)


anguta (approved):
looks good.

PR #375, opened 2020-09-22
fix bug where abscal switches are being skipped.

anguta:
a line was left out (only for abscal annoyingly) that reassigns the list of switches that tell `xrfi_run_step` whether to run or skip filters.


jsdillon (approved):
In theory, it would be nice to add some tests to expose this kind of issue, but I recognize that it's annoying. I approve right now, but perhaps make an issue if you agree @anguta

PR #377, opened 2020-10-27
Update XRFI day_threshold_run to properly handle a priori flags

jsdillon:
Right now, the weird interplay of masked arrays and nanmedians makes day_threshold_run behave very badly (right) when large sections of the band are flagged a priori. This PR removes masked arrays in favor of nanmedians, which behaves much better (left).

image


anguta (approved):
looks good.

PR #379, opened 2020-11-05
Fix filter_switches in XRFI where means and medians got interchanged

jsdillon:


anguta (approved):
Thanks for catching this. This looks good to me!

PR #380, opened 2020-11-07
Day threshold yaml

anguta:
no longer apply apriori freq and time flags in xrfi run (by default but include commandline arg for this). Add option to apply apriori flags post day thresholding.

### DIFF
--- a/hera_qm/utils.py
+++ b/hera_qm/utils.py
@@ -10,6 +10,7 @@ import argparse
 import numpy as np
 from pyuvdata import UVData
 from pyuvdata import UVCal
+from pyuvdata import UVFlag
 from pyuvdata import utils as uvutils
 from . import metrics_io
 from pyuvdata.telescopes import KNOWN_TELESCOPES
@@ -264,7 +265,7 @@ def get_metrics_ArgumentParser(method_name):
         ap.add_argument('--a_priori_flag_yaml', default=None, type=str,
                         help=('Path to a priori flagging YAML with frequency, time, and/or '
                               'antenna flagsfor parsable by hera_qm.metrics_io.read_a_priori_*_flags()'))
-        ap.add_argument('--a_apriori_ants_only', default=False, action="store_true",
+        ap.add_argument('--a_apriori_times_and_freqs', default=False, action="store_true",
                         help="If True, ignore frequeny and time flags in apriori yaml file (only use ant flags).")
         ap.add_argument('--xrfi_path', default='', type=str,
                         help='Path to save flag files to. Default is same directory as input file.')
@@ -676,10 +677,10 @@ def apply_yaml_flags(uv, a_priori_flag_yaml, lat_lon_alt_degrees=None, telescope
             input uvdata / uvcal but now with flags applied
     """
     # check that uv is UVData or UVCal
-    if not issubclass(uv.__class__, (UVData, UVCal)):
-        raise NotImplementedError("uv must be a UVData or UVCal object.")
+    if not issubclass(uv.__class__, (UVData, UVCal, UVFlag)):
+        raise NotImplementedError("uv must be a UVData, UVCal, or UVFlag object.")
     # only support single spw right now.
-    if uv.Nspws > 1:
+    if issubclass(uv.__class__, (UVData, UVCal)) and uv.Nspws > 1:
         raise NotImplementedError("apply_yaml_flags does not support multiple spws at this time.")
     # if UVCal provided, get lst_array from times.
     # If lat_lon_alt is not specified, try to infer it from the telescope name, which calfits files generally carry around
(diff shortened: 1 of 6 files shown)


hera_qm/xrfi.py
```diff
@@ -1398,6 +1398,7 @@ def xrfi_run_step(uv_files=None, uv=None, uvf_apriori=None,
                    dtype='uvcal', run_filter=True,
                    metrics=None, flags=None, modified_z_score=False,
                    a_priori_flag_yaml=None,
+                   a_priori_ants_only=False,
```
jsdillon, line 1401: I'd suggest making this default True, since we know that it can create problems when its False.

hera_qm/xrfi.py
```diff
@@ -1554,7 +1557,9 @@ def xrfi_run_step(uv_files=None, uv=None, uvf_apriori=None,
                 # No partial i/o for uvcal yet.
                 uv.read_calfits(uv_files)
                 if a_priori_flag_yaml is not None:
-                    uv = qm_utils.apply_yaml_flags(uv, a_priori_flag_yaml)
+                    uv = qm_utils.apply_yaml_flags(uv, a_priori_flag_yaml,
```
jsdillon, line 1560: Is there a reason this only gets applied to uvcal objects, but not uvdata objects?

hera_qm/xrfi.py
```diff
@@ -1571,7 +1576,9 @@ def xrfi_run_step(uv_files=None, uv=None, uvf_apriori=None,
                 else:
                     uv.read_calfits(uv_files)
                     if a_priori_flag_yaml is not None:
-                        uv = qm_utils.apply_yaml_flags(uv, a_priori_flag_yaml)
+                        uv = qm_utils.apply_yaml_flags(uv, a_priori_flag_yaml,
```
jsdillon, line 1579: Is there a reason this only gets applied to uvcal objects, but not uvdata objects?

hera_qm/xrfi.py
```diff
@@ -1696,6 +1705,7 @@ def xrfi_run_step(uv_files=None, uv=None, uvf_apriori=None,
 
 def xrfi_run(ocalfits_files=None, acalfits_files=None, model_files=None,
              data_files=None, a_priori_flag_yaml=None,
+             a_priori_ants_only=False,
```
jsdillon, line 1708: Again, I'd suggest default True

hera_qm/utils.py
```diff
@@ -264,6 +265,8 @@ def get_metrics_ArgumentParser(method_name):
         ap.add_argument('--a_priori_flag_yaml', default=None, type=str,
                         help=('Path to a priori flagging YAML with frequency, time, and/or '
                               'antenna flagsfor parsable by hera_qm.metrics_io.read_a_priori_*_flags()'))
+        ap.add_argument('--a_apriori_times_and_freqs', default=False, action="store_true",
```
jsdillon, line 268: Looks like you need to add YAML flags to day threshold run argparser.

scripts/xrfi_run.py
```diff
@@ -13,6 +13,7 @@
 
 xrfi.xrfi_run(args.ocalfits_files, args.acalfits_files, args.model_files,
               args.data_files, a_priori_flag_yaml=args.a_priori_flag_yaml,
+              a_priori_ants_only=not(args.a_apriori_times_and_freqs),
```
jsdillon, line 16: Looks like you need to add YAML flags to day threshold run script.

hera_qm/utils.py
```diff
@@ -691,62 +703,72 @@ def apply_yaml_flags(uv, a_priori_flag_yaml, lat_lon_alt_degrees=None, telescope
...
+            pol_array = uv.polarization_array
         elif issubclass(uv.__class__, UVCal):
-            ant_selection = uv.ant_array == antnum
-            if np.any(ant_selection):
-                for antind in np.where(ant_selection)[0]:
-                    uv.flag_array[antind, :, :, :, pol_selection] =True
+            pol_array = uv.jones_array
+        elif issubclass(uv.__class__, UVFlag):
+            pol_array = uv.polarization_array
+
```
jsdillon, line 751: I don't think you quite finished here: You need to handle the cases where it's a baseline time, an antenna type, and a waterfall type (which should do nothing).

hera_qm/utils.py
```diff
@@ -630,7 +633,8 @@ def strip_extension(path, return_ext=False):
 
 
 def apply_yaml_flags(uv, a_priori_flag_yaml, lat_lon_alt_degrees=None, telescope_name=None,
-                     ant_indices_only=False, by_ant_pol=False, ant_pols=None):
+                     ant_indices_only=False, by_ant_pol=False, ant_pols=None,
```
jsdillon, line 636: I'd like to see the unit tests of this function updated to include these new options.

jsdillon (changes requested):
Thanks for the quick turn-around on this @anguta! Looking good so far. There's a handful of minor tweaks and a few features / cases that I think are missing.

hera_qm/utils.py
anguta: It seems that the polarization_array is always there, even if the uvflag is antenna type (despite the fact that uvcal objects use jones_array instead of polarization_array). Here is the line where the assignment happens.

https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/eedb1e10ef2af8e4a019e7df2abed012d23c6471/pyuvdata/uvflag/uvflag.py#L3042



 If the type is a waterfall, the the polarization array contains a string so that no pols will come up as being contained in the array so nothing should happen.

scripts/xrfi_run.py
anguta: I've added this. Thanks for catching!

hera_qm/utils.py
anguta: Thanks for catching this. I've updated the parser.

hera_qm/xrfi.py
anguta: It also gets applied to uvdata objects but because these lines only read the metadata, it doesn't happen until here 1601 when the data is actually read in.

hera_qm/xrfi.py
anguta: It also gets applied to uvdata objects but because these lines only read the metadata, it doesn't happen until here 1601 when the data is actually read in.

hera_qm/xrfi.py
anguta: These need to have default = False to preserve many of the unittests. The script interace defaults everything to True though.

jsdillon: I'm fine with this.

hera_qm/xrfi.py
anguta: I'm leaving the default to False because this behavior is less confusing and the problems being created are not inherent to apriori flags but rather their interaction with the median filtering strategy (in a way that should be similar to edge effects?). 

I have added a warning though that spurious over-flagging can occur within a kernel width of the flags.

jsdillon (approved):
Looks good. Thanks @anguta!

PR #383, opened 2020-11-10
Correctly apply apriori flags to waterfall uvflag objects

anguta:
waterfall uvflag objects have freq_arrays with shape (Nfreq) rather then (Nspw, Nfreq). This was not be treated correctly by apply_apriori_flag_yaml.


jsdillon (approved):
Approved, pending tests passing. I'd like to pull this in ASAP, but this error indicates to me that our unit tests didn't expose the bug. Can we make an issue to expand the unit tests to cover this case (or just another PR to expand those tests)?

anguta:
> Approved, pending tests passing. I'd like to pull this in ASAP, but this error indicates to me that our unit tests didn't expose the bug. Can we make an issue to expand the unit tests to cover this case (or just another PR to expand those tests)?

I've filed an issue.

PR #386, opened 2020-11-30
Make night_index work when uvlist is relative path

jsdillon:
I was running into a problem with H1C IDR3.2 where `all_files` are absolute paths but `uvlist` is all relative paths. This breaks `night_index = all_files.index(os.path.abspath(uvlist[ind]))`, so this fixes that.


anguta (approved):
I approve, pending the tests passing.

PR #387, opened 2020-12-12
Fix issues related to new UVCal metadata and warnings

bhazelton:
UVCal was just updated to include new metadata and a few warnings associated with that were added. This update broke one line of code in `utils` that assumed that the `lst_array` was either not an attribute or required, which is not always true for UVCal (it is currently optional but will become required in a future version of pyuvdata). It also broke some tests because new warnings were being issued.

I added some pytest warning filters to filter a bunch of pyuvdata warnings that I do not think are too important for hera_qm, but someone with more hera_qm knowledge should check that.

### DIFF
--- a/hera_qm/tests/test_ant_metrics.py
+++ b/hera_qm/tests/test_ant_metrics.py
@@ -11,6 +11,9 @@ from hera_qm import metrics_io
 from hera_qm.data import DATA_PATH
 import hera_qm.tests as qmtest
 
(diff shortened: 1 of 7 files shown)


bhazelton:
@anguta It looks like you were the primary author of the code and test I had to do the most changes on. Can you review?

jsdillon (approved):
Looks good to me!

PR #390, opened 2021-01-15
Remove center pixel from medfilt footprint

jsdillon:
When performing `detrend_medfilt` on a very smooth function (i.e. with a near-constant slope and little noise), I was getting a lot of modified z-scores of `np.inf`. This arose because the median pixel in the kernel was often the pixel of interest. This led to residuals of 0. With high enough density of 0 residuals, doing the median filter of the residual squared to estimate the denominator of the modified z-score led to a lot of zeros in the denominator and thus infinite z-scores. This wasn't a problem for most XRFI tasks in the current pipeline, which were noisy and also had both real and imaginary parts that both had to match, but I was recently trying to find RFI in H4C antenna-averaged autocorrelations (which is very low noise) and it was flagging large RFI-free regions as a result of these `inf`s.

This PR removes the pixel of interest from the median calculation using `scipy.ndimage.median_filter`'s `footprint` kwarg (we moved to using that function in #389, since it's slightly faster and I was getting weird segfaults with the old one). This makes it far less likely that the residual is 0 in a data with any kind of noise. 

Preliminary tests indicate that it has no impact on runtime.


adampbeardsley:
I like this solution Josh. Could you add a unit test, or paste in some plots showing the solution fixes the issue?

jsdillon:
Here's the waterfall I'm looking at (it's the anntenna averaged autocorrelation): 

image

Here's the metric before and after: 
image

The white is where the old metric is `np.inf`.

adampbeardsley (approved):
Thanks, looks good to me.

PR #391, opened 2021-01-23
New Module auto_metrics uses day-long waterfalls to find outlier antennas

jsdillon:
This PR creates a new module, `auto_metrics`, designed to run on day-long waterfalls of autocorrelations to look for outliers. In particular, we are looking for outliers in either shape, power, temporal variability, or temporal discontinuities. In general, these are assessed by collapsing each waterfall to a single (normalized) spectrum, then comparing that spectrum to the mean/median of good spectra to compute some difference metric. Those are then converted into modified Z-scores by comparing the overall distribution of good antennas. This whole processes is repeated iteratively until no new bad antennas are identified.

All of that happens twice, first with more robust median-based statistics, then (after a round of very simple RFI flagging) with more sensitive mean-based statistics. 

### Statistics Computed:
- **Shape:**
   - `median_spectra_normed`: Compute the median spectrum over time for each autocorrelation, but then divide by the median of the whole waterfall to get something near 1. Now compute the median absolute value of the difference between that and the median spectrum over all non-excluded antennas of the same polarization. Convert this to a modified Z-score by comparing to all non-excluded antennas.
   - `mean_spectra_normed`: Same as above, but replace medians with means (still use a modified Z-score at the end, though).
- **Temporal variability:** 
   - `mad_spectra_normed`: Divide each autocorrelation by the median waterfall for all non-excluded antennas of the same polarization. Now reduce to a single spectrum by computing the MAD along the time axis for each. Now compute the median value (not the absolute value, since low variability shouldn't get a high z-score) of the difference between that and the median spectrum over all non-excluded antennas of the same polarization. Convert this to a modified Z-score by comparing to all non-excluded antennas.
   - `std_spectra_normed`: Same as above, but the MAD becomes a standard deviation and the medians become means (still use a modified Z-score at the end, though). 
- **Temporal discontinuities:** 
   - `median_abs_diff_spectra_normed`: Divide each autocorrelation by the median waterfall for all non-excluded antennas of the same polarization. Now compute the element-by-element diff along the time axis, take the absolute value, and take the median to get a single spectrum. Now compute the median value (not the absolute value, since low variability shouldn't get a high z-score) of the difference between that and the median spectrum over all non-excluded antennas of the same polarization. Convert this to a modified Z-score by comparing to all non-excluded antennas.
   - `mean_abs_diff_spectra_normed`: Same as above, but replace medians with means (still use a modified Z-score at the end, though).
- **Power:** 
   - `median_spectra`: Compute the median spectrum over time for each autocorrelation (no normalization). Now compute the median absolute value of the difference between the log of that (because we care about power outliers in dB rather than linear units) and the log of the median spectrum over all non-excluded antennas of the same polarization. Convert this to a modified Z-score by comparing to all non-excluded antennas.
   - `mean_spectra`: Same as above, but replace medians with means (still use a modified Z-score at the end, though).

This code runs in about 30 minutes on a full day of H4C data (2459122), assuming that the data have already gone through a round of autocorrelation extraction. If not, the I/O (which is already nearly 1/3 of that time) is a lot slower.

Here's some example results from that day. Here I show all four metrics' spectra (red if it was flagged for that reason, purple if flagged for another reason) compared to the grey spectra of nominally good antennas. I also show the waterfall (after basic RFI flagging) and the log of the waterfall after dividing out the average good waterfall for that polarization. 

Screen Shot 2021-01-25 at 7 25 25 PM

While there may be some fine tuning that needs to happen to figure out precisely what we think "bad" means, this should be a much more sensitive way of identifying bad antennas via autocorrelations. Hopefully, some combination of this (with an accompanying notebook), @dstorer's forthcoming update of `ant_metrics` to use the correlation matrices currently in the `data_inspect` notebooks, and chi^2 per antenna, we will be in good shape for the next observing campaign.


hera_qm/auto_metrics.py
```diff
@@ -0,0 +1,412 @@
...
+
+
+def nanmad(a, axis=None):
+    '''Analogous to np.nanmedian, but for median absolute deviation.'''
+    nanmed = np.nanmedian(a, axis=axis, keepdims=True)
+    return np.nanmedian(np.abs(a - nanmed), axis=axis)
+
+
+def nanmedian_abs_diff(a, axis=0):
+    '''Computes the absolute difference between neightbors along axis, then collaposes
```
dstorer, line 21:
```suggestion
    '''Computes the absolute difference between neighbors along axis, then collapses 
```

hera_qm/auto_metrics.py
```diff
@@ -0,0 +1,412 @@
...
+
+
+def nanmedian_abs_diff(a, axis=0):
+    '''Computes the absolute difference between neightbors along axis, then collaposes 
+    along that axis with the nanmedian. Useful for studying temporal variability, e.g.'''
+    return np.nanmedian(np.abs(np.diff(a, axis=axis)), axis=axis)
+
+
+def nanmean_abs_diff(a, axis=0):
+    '''Computes the absolute difference between neightbors along axis, then collaposes
```
dstorer, line 27:
```suggestion
    '''Computes the absolute difference between neighbors along axis, then collapses 
```

hera_qm/xrfi.py
```diff
@@ -1053,6 +1053,70 @@ def flag_apply(uvf, uv, keep_existing=True, force_pol=False, history='',
...
+    flags |= _ws_flag_waterfall(meanfilt_metric, flags, nsig=sig_adj)
+    
+    # Perform spectral thresholding 
+    combined_metric = np.where(medfilt_flags, medfilt_metric, meanfilt_metric)
+    spec = np.nanmedian(combined_metric, axis=0)
+    zspec = modzscore_1d(spec, detrend=False)
+    zspec[~np.isfinite(zspec)] = sig_init
+    flags[:, _ws_flag_waterfall(np.abs(zspec), np.abs(zspec) >= sig_init, nsig=sig_adj)] = True
+
+    # Perform temporal thresholding
```
dstorer, line 1101: What sort of temporal structure do we expect to show up in the averaged waterfall? This concerns me a little, because if it's leaking in from the temporal instability of a particular antenna, even if we flag the integrations that contain the discontinuity, there may be residual bandpass weirdness in integrations before/after the actual discontinuity. This isn't necessarily a criticism of flagging temporal structure, but more a commentary that if we have to flag temporal structure it means we are getting more of it in the averaged waterfall than we would hope, since dividing out a waterfall with that type of structure might explain why we were seeing so much temporal structure in your waterfalls. I'm not sure I have a specific fix in mind, mostly just wondering if you've already thought about this, or if I'm interpreting things wrong.

dstorer (approved):
This all looks good, my only concern that potentially involves changes is regarding the temporal structure in the mean waterfall - see my previous comment for those thoughts and let me know what you think. I'm also okay with proceeding as is and crossing that bridge once the current implementation is set and working.

hera_qm/xrfi.py
jsdillon: This is a good question. If all antennas have a temporal discontinuity at the same time in the same way, this algorithm will attribute that to RFI rather than an antenna problem. When an integration is flagged, it's ignored in the collapsed spectra/metrics (and in a diff-based metric, np.diff will propagate to the nan to both diffs involving that pixel). I think that's consistent with what we're seeing, which is apparently uncorrelated temporal discontinuities between antennas. Also, this flagging algorithm is applied only after throwing out the most egregious antennas in the "median round," so the remainder shouldn't add much to the rumble of the antenna-averaged waterfall that we then flag on.

A separate question is whether temporal structure in one antenna, when we average all antennas together to produce a per-polarization waterfall to normalize out for the "Temporal Variability" and "Temporal Discontinuity" metrics, affects the measured temporal structure in the others. We expect that it should, at some level, but that's why I have these lines:
https://github.com/HERA-Team/hera_qm/blob/2192099f6964a52d2a86fa18dc33c7a8f788aacb/hera_qm/auto_metrics.py#L388-L392

After each round of flagging, I recompute the spectra that depend on waterfall normalization, removing them from the average, before re-running the Z-score calculation. This should help reduce the cross-contamination of bad antennas.

dstorer: This pretty much answers my question, and I think the way you've implemented things minimizes "structure leakage" well (and I can't think of a straightforward change that would make it better). One of the reasons for my concern is that when looking at the SSINS of this data we were seeing a lot of temporal structure, and SSINS does a mean over all baselines after the time-differencing steps, and I think even though the temporal structure isn't aligned between antennas it was still bright enough to show up in the mean even when it was only present in one antenna. It should affect all antennas roughly the same, so I don't think this will really affect which antennas we flag, but I do think it could produce a bias where the worst antennas look even worse because they are being compared against an average waterfall with structure. I'm fine with moving forward with things as they are for now though, and just keeping an eye on whether this seems to be causing problems.

dstorer (approved):

PR #392, opened 2021-01-29
Minor bugfix in auto_metrics

jsdillon:
A line that should only get triggers when there are multiple input files was getting triggered for 1 input auto file.


dstorer (approved):
All good!

PR #394, opened 2021-02-02
Fix definition of modified z-score in auto_metrics

jsdillon:
I was multiplying by 1.4826 instead of dividing by 1.4826. So all the modified z-scores in previous calculations were off by a factor of ~2. This fixes that bug and changes the defaults to reflect that change.


dstorer (dismissed):
Looks good

dstorer (dismissed):

PR #396, opened 2021-02-26
Add support for multiple metrics files, including auto_metrics, to XRFI

jsdillon:
This is basically the same as https://github.com/HERA-Team/hera_cal/pull/687


dstorer (approved):
Looks good!

PR #398, opened 2021-03-02
discard data associated with antennas flagged in yaml file.

anguta:
New functionality in `apply_yaml_flags` to throw away data associated with flagged antennas to save us space and time.


jsdillon (approved):
Looks good to me.

PR #399, opened 2021-03-11
Fix bug in process_ex_ants

jsdillon:
@dstorer noticed an error in the logic of process_ex_ants where if `ex_ants` is not specified but `metrics_files`, the result is just an empty list.

This fixes that bug and exposes the error in a unit test.


dstorer (approved):
Looks good, thanks for fixing this.

PR #400, opened 2021-03-12
add unflag_first option to yaml flags.

anguta:
add an option to apply_yaml_flags to unflag the uvdata/uvcal/uvflag object before applying flags.

### DIFF
--- a/hera_qm/tests/test_utils.py
+++ b/hera_qm/tests/test_utils.py
@@ -369,7 +369,7 @@ def test_apply_yaml_flags_uvcal(filein, new_metadata):
     channel_flags = [0, 1, 60] + list(range(10, 21)) # channels from yaml file.
     integration_flags = [0, 1] # integrations from yaml file that should be flagged.
     ant_flags = [0, 10, [1, 'Jee'], [3, 'Jnn']]
-    if 'no_flags' not in test_flag:
+    if 'no_flags' in test_flag:
         # check that the uvcal is completely unflagged since we used unflag_first
         assert not np.any(uvc.flag_array)
     else:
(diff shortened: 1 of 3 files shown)


hera_qm/utils.py
```diff
@@ -722,6 +726,8 @@ def apply_yaml_flags(uv, a_priori_flag_yaml, lat_lon_alt_degrees=None, telescope
         raise NotImplementedError("apply_yaml_flags does not support multiple spws at this time.")
     # if UVCal provided and lst_array is None, get lst_array from times.
     # If lat_lon_alt is not specified, try to infer it from the telescope name, which calfits files generally carry around
+    if unflag_first:
+        uv.flag_array[:] = False
```
jsdillon, line 730: I think this is fine for any dimensions, but I'd recommend `uv.flag_array = np.zeros_like(uv.flag_array, dtype=bool)` to be more explicit.

hera_qm/tests/test_utils.py
```diff
@@ -369,7 +363,7 @@ def test_apply_yaml_flags_uvcal(filein, new_metadata):
         uvc.antenna_positions = None
         uvc.lst_array = None
 
-    uvc = utils.apply_yaml_flags(uvc, test_flag)
+    uvc = utils.apply_yaml_flags(uvc, test_flag, unflag_first=True)
```
jsdillon, line 366: Do you actually want to show that previous flags were removed in this test?

jsdillon (changes requested):
Couple of minor suggestions.

hera_qm/utils.py
anguta: Done.

hera_qm/tests/test_utils.py
anguta: I've added a check.

anguta:
> Couple of minor suggestions.

Thanks for the feedback. Should be good to go (pending tests passing 🤞 )

jsdillon (approved):

PR #401, opened 2021-06-20
Make fixes for numpy v1.20

plaplant:
In the upgrade to numpy v1.20, [several datatypes like `np.int` were formally deprecated](https://numpy.org/doc/stable/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated). This led to the incorrect number of warnings being raised for some hera_qm tests, so they were failing. This PR removes references to the deprecated types, and cleans up some of the tests in the process.

### DIFF
--- a/hera_qm/metrics_io.py
+++ b/hera_qm/metrics_io.py
@@ -116,7 +116,7 @@ def _recursively_save_dict_to_group(h5file, path, in_dict):
         with h5py, a TypeError is raised.
 
     """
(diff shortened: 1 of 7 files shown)


mkolopanis (approved):
This all looks pretty good to me. Got to be defensive against the 1.20 change.

Codecov is complaining about a line that has literally never been covered. I'm wondering why that line even exists to begin with because the error requires data that doesn't exist to try to be taken from an object.

PR #404, opened 2021-12-01
Break file loading in ant_metrics into its own function to prevent memory leaks

jsdillon:
Right now, ant_metrics is taking more than 128 GB of memory when running on 10 files and appears to be leaking memory quite badly:

image

This PR breaks file loading into a function, which is looped over per files group, to help make sure everything goes out of scope as appropriate.


jsdillon:
And here's the memory profile on the same files with this branch:

image

dstorer (approved):
This looks good - thanks for the fix.

PR #405, opened 2021-12-10
Enable faster process_ex_ants using h5py directly

jsdillon:
`process_ex_ants` doesn't actually need to load entire metrics files, which can be quite slow when the metrics files are huge. This PR tries to use h5py to just extract the part of the file we actually need.

### DIFF
--- a/hera_qm/tests/test_metrics_io.py
+++ b/hera_qm/tests/test_metrics_io.py
@@ -465,11 +465,12 @@ def test_process_ex_ants_bad_string():
 
 def test_process_ex_ants_string_and_file():
     ex_ants = '0,1'
-    met_file = os.path.join(DATA_PATH, 'example_ant_metrics.hdf5')
-    xants = metrics_io.process_ex_ants(metrics_files=met_file)
-    assert xants == [81]
-    xants = metrics_io.process_ex_ants(ex_ants=ex_ants, metrics_files=met_file)
-    assert xants == [0, 1, 81]
+    for ext in ['hdf5', 'json']:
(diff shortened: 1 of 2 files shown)


hera_qm/tests/test_metrics_io.py
```diff
@@ -465,11 +465,12 @@ def test_process_ex_ants_bad_string():
 
 def test_process_ex_ants_string_and_file():
     ex_ants = '0,1'
-    met_file = os.path.join(DATA_PATH, 'example_ant_metrics.hdf5')
-    xants = metrics_io.process_ex_ants(metrics_files=met_file)
-    assert xants == [81]
-    xants = metrics_io.process_ex_ants(ex_ants=ex_ants, metrics_files=met_file)
-    assert xants == [0, 1, 81]
+    for ext in ['hdf5', 'json']:
```
plaplant, line 468: Instead of doing this as a `for` loop inside of the test, we could pretty easily make this into two separate tests using [pytest's parametrize functionality](https://docs.pytest.org/en/6.2.x/parametrize.html), which is relatively easy to implement. I think that's better testing practice because if one test fails but not the other, it makes it more obvious where the problem is coming from.

plaplant (commented):
Looks great, thanks @jsdillon! I just have a small comment about the change to the tests.

hera_qm/tests/test_metrics_io.py
jsdillon: Good idea! Done.

plaplant (approved):
Thanks! It looks like we're missing a few lines of coverage which we should probably cover, but I'm okay with merging it in to speed things up on site.

PR #406, opened 2021-12-11
Add option to ignore cross-polarized visibilities in XRFI

jsdillon:
I've been investigating how to make XRFI run more quickly on site.

I was under the impression for a while that this was actually how XRFI worked when looking at only autos, though I just figured out that that's not the case.

This PR leaves the status quo as the default. Whether we want to use this option in future stage 2 XRFI is an open question.


aewallwi (changes requested):
We should add a unittest for the skip_cross_pol_vis option but I'm happy with the implementation.

aewallwi (approved):
Looks good! I've approved.

PR #407, opened 2022-01-20
Only load uvdata in xrfi_run_step when run_filter is true

jsdillon:
@ele-rath and I were trying to debug the weird extra flagging sideband feature in XRFI and we noticed that XRFI was loading data at this step https://github.com/HERA-Team/hera_qm/blob/d5d91cd9b768b99faeed3e7819ce128639d09330/hera_qm/xrfi.py#L1670 even when `run_filter` is False. I can't tell if this is a bug or a design choice. @aewallwi ???


jsdillon:
I'm closing this in favor of #408 which also achieves the same goals

PR #408, opened 2022-01-22
Rewrite xrfi_run more cleanly and concisely

jsdillon:
By changing what's looped over and what's explicitly written out and using a dictionary for repeated kwargs, I've vastly simplified `xrfi_run` without changing its inputs or outputs. Hopefully, this will make things easier to debug.

### DIFF
--- a/hera_qm/xrfi.py
+++ b/hera_qm/xrfi.py
@@ -1973,22 +1973,20 @@ def xrfi_run(ocalfits_files=None, acalfits_files=None, model_files=None,
     """
     if ocalfits_files is None and acalfits_files is None and model_files is None and data_files is None:
         raise ValueError("Must provide at least one of the following; ocalfits_files, acalfits_files, model_files, data_files")
-    # user must provide an optional output prefix if no data file is provided.
-    if isinstance(acalfits_files, (str, np.string_)):
-        acalfits_files = [acalfits_files]
-    if isinstance(ocalfits_files, (str, np.string_)):
-        ocalfits_files = [ocalfits_files]
-    if isinstance(model_files, (str, np.string_)):
-        model_files = [model_files]
-    if isinstance(data_files, (str, np.string_)):
-        data_files = [data_files]
-    if isinstance(output_prefixes, (str, np.string_)):
-        output_prefixes = [output_prefixes]
+
+    # recast strings as lists of strings
+    for varname in ['acalfits_files', 'ocalfits_files', 'model_files', 'data_files', 'output_prefixes']:
+        if isinstance(eval(varname), (str, np.string_)):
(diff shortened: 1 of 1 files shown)


aewallwi (changes requested):
Make sure to perform a comparitive test to ensure that the flags and metrics from the rewritten version of this function match the flags from the old version of this function.

jsdillon:
image

I've run XRFI on both this branch and the master branch for a subset of 2459122 with the same settings used for that day. I get the same results (except in the history field, of course), so I'm feeling pretty confident I haven't introduced any bugs.

hera_qm/xrfi.py
```diff
@@ -1998,254 +2003,154 @@ def xrfi_run(ocalfits_files=None, acalfits_files=None, model_files=None,
...
-        if len(flags) > 1:
-            for flg in flags[1:]:
-                vdict['uvf_init'] |= flg
-        vdict['uvf_init'].label = 'ORd flags, round 1.'
 
+    def _run_all_filters(median_round, 
+                         omnical_filter, omnical_chi2_filter, omnical_zscore_filter, 
+                         abscal_filter, abscal_chi2_filter, abscal_zscore_filter,
+                         omnivis_filter, cross_filter, auto_filter):
+        '''This function runs all possible filters, updating vdict as appropriate'''
```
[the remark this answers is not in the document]
aewallwi: Doesn't need to be a full fledged one, but just a sentence describing the nature of all the args (they are switches to run the different types of filter/flag step).

hera_qm/xrfi.py
```diff
@@ -1998,254 +2003,154 @@ def xrfi_run(ocalfits_files=None, acalfits_files=None, model_files=None,
...
+        
+        # Start with empty list of flags and metrics for later combination
+        metrics = []
+        flags = []
+        
+        # Modify these labels and parameters based on whether we're doing mean- or median-based statistics
+        label = {True: 'median', False: 'mean'}[median_round]
```
aewallwi, line 2031: note the True/False nature of median round in the docstring or use a stringe for median_round

aewallwi (changes requested):
Generally looking pretty good. I think this yielded substantial readability and maintainability improvements. Just a few minor adjustments still. A docstring clarifications in _run_all_filters would be nice. Also, I think it'd be good to check that you get the exact same results from the old unittests in `test_xrfi_run` to the new version -- not that the old one is any more correct but just to make sure we are making controlled changes (preserving the exact old functionality).

hera_qm/xrfi.py
jsdillon: Done.

aewallwi (approved):
Looks good. I approve.

PR #409, opened 2022-01-29
Add time_atol to flag_apply to fix XRFI overflagging

jsdillon:
@ele-rath and I discovered that one major source of over-flagging in XRFI, as discussed in section 6 of http://reionization.org/manual_uploads/HERA097_H1C_IDR3_2_Memo.pdf, was a bug in XRFI where UVFlag objects created from UVCal objects could not be properly applied to UVData objects when the times were slightly mismatched (an issues we've encountered before).

In our tests, we see rounding errors cause certain integrations identified as bad during the median filter round of XRFI to not get flagged prior to the mean filter round because times were mismatched between the data files and calfits files. In the case of our tests on 2459122, the mismatch was at the ~40 microsecond level. This then led to very bright RFI skewing the statistics of the mean filter and creating edge-effects based on the size of the kernel.

This PR uses the new `time_atol` parameter in https://github.com/RadioAstronomySoftwareGroup/pyuvdata/pull/1133 to allow times that are close but not perfectly matching to get identified as the same. 

Here's a flagging pattern for a subset of that day before this fix:

image

And here's the same pattern after this fix:

image

This PR builds off #408 and requires approval of https://github.com/RadioAstronomySoftwareGroup/pyuvdata/pull/1133 in order for tests to pass.

### DIFF
--- a/hera_qm/xrfi.py
+++ b/hera_qm/xrfi.py
@@ -2021,7 +2021,8 @@ def xrfi_run(ocalfits_files=None, acalfits_files=None, model_files=None,
                          omnical_filter, omnical_chi2_filter, omnical_zscore_filter, 
                          abscal_filter, abscal_chi2_filter, abscal_zscore_filter,
                          omnivis_filter, cross_filter, auto_filter):
-        '''This function runs all possible filters, updating vdict as appropriate'''
+        '''This function runs all possible filters, updating vdict as appropriate. 
+        rnd_label is "mean" or "median". All other args are booleans for whether to run that filter type.'''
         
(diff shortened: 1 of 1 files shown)


jsdillon:
It appears to be the case that instead of implementing a new parameter in pyuvdata, we're instead using the `rtol` and `atol` built into the `_time_array` parameter. In that case, this PR is no longer necessary, since running on the master branch of hera_qm (and the branch of pyuvdata in https://github.com/RadioAstronomySoftwareGroup/pyuvdata/pull/1133) produces the same result:

image

PR #410, opened 2022-01-31
Add option to ignore xants in xrfi_run_step to use on redundantly-averaged visibilities or omnical visibility solutions

jsdillon:
This PR adds an option to xrfi_run_step to completely override any a priori antenna flagging. This is necessary for omnical visibility solutions, which are already averaged over redundant baselines and thus keyed by potentially flagged antennas. This prevents redundant baseline groups from getting ignored in flagging unnecessarily.


ele-rath (approved):
Everything looks good to me

PR #411, opened 2022-02-13
Allow median filter and mean filter rounds in XRFI to flag on different numbers of sigma

jsdillon:
In theory, the two rounds in XRFI were meant to catch the most egregious outliers with the medfilt round and then follow-up with the more sensitive meanfilt round.

In practice, I was noticing that the median filter was noisier and leading to more spurious flags. Inspired by the approach that auto_metrics takes, where the cut on median statistics was more lenient than the subsequent mean-based cut, I've enabled separate values of `nsig_init` and `nsig_adj` for the mean and median rounds.

Here's where they are both 5 and 2 sigma: 

image

And here's where it's 10 and 4 sigma for the median round and then 5 and 2 sigma for the mean round:

image

(Ignore changes at the band edges, because I also varied some a priori flags between these tests.) This definitely seems helpful and I feel that if the extra points were not caught by the mean filter, they aren't real RFI.


ele-rath (approved):
Something to check: does raising sig_init_median and sig_adj_median still capture all of the known channels with RFI at all times (ex: at 62.5 MHz from the clock)? Since I have already checked this, I will approve

PR #412, opened 2022-04-19
Introduce alternate absolute amplitude z-scores in auto_metrics

jsdillon:


hera_qm/auto_metrics.py
```diff
@@ -118,7 +118,11 @@ def spectrum_modz_scores(auto_spectra, ex_ants=[], overall_spec_func=np.nanmedia
         If True, take the log of both the spectrum and the overall spectrum before taking the abs diff
     abs_diff : bool
         If True, the metric uses the abs diff with the overall spectrum. Otherwise, it's just the diff.
-
+    minimum_relative_sigma : float
```
dannyjacobs, line 121: This explanation is hard to follow. Let me see if I can offer an alternate translation.
The design of the analog system allows gain to vary by at least 5dB without loss of linearity. This option limits the standard deviation used in calculation of the amplitude zscore to a fractional amount of the autocorrelation amplitude.  For example a value of 0.1 sets a lower bound of 1dB on the variation.  When paired with a zscore threshold of 5 this will have the effect of allowing 5dB of variation.  Adjust this in concert with the zscore to change the allowed gain variability.

jsdillon: @dannyjacobs I've re-written how this function is modified. I hope it's clearer, but I'm not 100% certain. I'm also not 100% certain this is how we want to impose this minimum, but I can do some more testing if you think the basic framework seems reasonable.

jsdillon:
This has been completely superceded by the new notebooks, so I'm closing this.

PR #413, opened 2022-05-24
Use future_array_shapes

jsdillon:
Following https://github.com/RadioAstronomySoftwareGroup/pyuvdata/pull/1186, this PR updates hera_qm to use the new pyuvdata shapes. This was particularly complicated for XRFI, which involved a bunch of array slicing and averaging. 

Once this is done, we should be able to do the same for hera_cal: https://github.com/HERA-Team/hera_cal/pull/797

I've tagged both @adampbeardsley and @bhazelton as reviewers. A lot of this was your code @adampbeardsley, so I think you're best suited to reviewing this, but I'm not sure if you're available to review a PR.


jsdillon:
@bhazelton @adampbeardsley OK, I've fixed my numpy version wrangling and this PR is ready for a review.

hera_qm/tests/test_xrfi.py
```diff
@@ -48,27 +48,30 @@
...
+    uvc.use_future_array_shapes()
     xant = uvc.ant_array[0]
     xrfi.flag_xants(uvc, xant)
-    assert np.all(uvc.flag_array[0, :, :, :, :])
+    assert np.all(uvc.flag_array[0, :, :, :])
 
 
 def test_uvflag():
     uvf = UVFlag(test_f_file)
+    uvf.use_future_array_shapes()
```
bhazelton, line 69: There's actually also a way to do this in one line, you can do:
`uvf = UVFlag(test_f_file, use_future_array_shapes=True)`

We're adding a similar option to UVData and UVCal read and from_file methods in an upcoming PR.

bhazelton (approved):
This looks good to me.

PR #416, opened 2022-06-08
Fix version handling and automated testing

jsdillon:
These are updates that should have been part of #415


steven-murray (approved):
This all looks right to me!

PR #419, opened 2022-07-29
Add `ant_class` module for classifying antennas based on absolute bounds for various metrics

jsdillon:
This PR adds a new module `ant_class` and the `AntennaClassification` object, which allows for the flexible categorization of ant-pol tuples. 

It is initialized like `AntennaClassification(good=[(0, 'Jee'), (1, 'Jee')], whatever=[(0, 'Jnn'), (10, 'Jee')])` and takes arbitrary categorizations. Allows setting and getting of a specific antenna's classification using `[]`, as well as `.ants` and `.classes` properties and a `.get_all(classification)` function.

However, there are three special "quality" classifications (by default `"good"`, `"suspect"`, and `"bad"`, but they can be modified) that are queried via the property e.g. `.good_ants` or the function `.is_good(ant).` Furthermore, these classifications the sensible enable combination of `AntennaClassification` objects by overloading `+` such that good + good = good, good + bad = bad, good + suspect = suspect, bad + suspect = bad, and bad + bad = bad.

This PR adds also a new function `antenna_bounds_checker()` to `ant_class` designed for turning scalar antenna metrics into `AntennaClassification` objects. This function is very general and I plan to add absolute limits checkers built on top of it.

This PR lays the groundwork for future absolute limits checking. Eventually, the goal here is to implement new absolute limit checking on antenna power, bandpass slope, etc. as has been prototyped by @AaronParsons in https://github.com/HERA-Team/hera_notebook_templates/blob/ef53071cfac322a6fe9d629d0ed508028bd1fe3c/notebooks/fastcal_inspect.ipynb

### DIFF
--- a/hera_qm/ant_class.py
+++ b/hera_qm/ant_class.py
@@ -197,10 +197,10 @@ def antenna_bounds_checker(data, **kwargs):
     Example usage: antenna_bounds_checker(data, good=[(.5, 1)], suspect=[(0, .5), (1, 2)], bad=(-np.inf, np.inf))
     
     Arguments:
-        data: dictionary mapping ant-pol tuples (or autocorrelation keys) to 
+        data: dictionary mapping ant-pol tuples (or autocorrelation keys) to scalar values
         kwargs: named antenna classifications and the range or ranges of data values which map to those 
-            categories. Classification bounds are checked in order and the first accepted bound is used. 
-            All ranges are inclusive (i.e. >= or <=).
+            categories. Classification bounds are checked in order and the first accepted bound is used, 
+            though it is possible for antenna to not get classified. All ranges are inclusive (i.e. >= or <=).
         
     Returns:
         AntennaClassification object using data and bounds to classify antennas in data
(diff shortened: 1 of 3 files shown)


hera_qm/ant_class.py
```diff
@@ -0,0 +1,239 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2022 the HERA Project
+# Licensed under the MIT License
+
+"""Class and algorithms to classify antennas by various data quality metrics."""
+import numpy as np
+
+
+def _check_antpol(ap):
+    '''Verifies that input is a valid ant-pol tuple. Otherwies, raises a ValueError'''
```
tyler-a-cox, line 10: Minor typo here "Otherwies"

tyler-a-cox (changes requested):
Looks good to me! I find a minor typo in one of the doc strings, but besides that I think this is really clear. Happy to approve after typo gets fixed

jsdillon:
Ok, fixed. Thanks @tyler-a-cox!

tyler-a-cox (approved):
Looks good! Thanks @jsdillon

PR #420, opened 2022-08-03
Ant metrics fast reader

jsdillon:
This PR migrates `ant_metrics` over to using the new `HERADataFastReader`, which should improve I/O by a factor of 2 or better.

This PR also gives advanced users the option of using the `AntennaMetrics` class with only minimal I/O by providing `DataContainer`s directly. This should be useful for interactive work, like the fastcal inspect notebook currently being developed: https://github.com/HERA-Team/hera_notebook_templates/blob/master/notebooks/fastcal_inspect.ipynb


tyler-a-cox (approved):
This looks good to me as well. Thanks for adding this!

PR #421, opened 2022-08-04
Add `__str__` method to `AntennaClassification`

jsdillon:
This allows the user to print an `AntennaClassification` object and get a easily read report of antenna classifications.


tyler-a-cox (approved):
This one seems pretty straightforward. Approved

PR #423, opened 2022-08-18
Add functions for flagging auto-correlations

tyler-a-cox:
This PR implements new RFI flagging functions for auto-correlation using channel differencing and DPSS filtering from @AaronParsons [fastcal notebook](https://github.com/HERA-Team/hera_notebook_templates/blob/ef53071cfac322a6fe9d629d0ed508028bd1fe3c/notebooks/fastcal_inspect.ipynb).

### DIFF
--- a/hera_qm/ant_class.py
+++ b/hera_qm/ant_class.py
@@ -6,6 +6,7 @@
 import numpy as np
 from scipy.ndimage import median_filter
 import warnings
+from . import xrfi
 
 
 def _check_antpol(ap):
@@ -320,3 +321,57 @@ def auto_slope_checker(data, good=(-.2, .2), suspect=(-.4, .4), edge_cut=100, fi
 
     return antenna_bounds_checker(relative_slopes, good=good, suspect=suspect, bad=(-np.inf, np.inf))
 
+def auto_rfi_checker(data, good=(0, 0.1), suspect=(0.1, 0.2), nsig=6, antenna_class=None, flag_threshold=0.1, 
+                     kernel_widths=[3, 4, 5], mode='dpss_matrix', filter_centers=[0, -2700e-9, 2700e-9],
+                     filter_half_widths=[200e-9, 200e-9, 200e-9], eigenval_cutoff=[1e-9]):
+    """
+    Classifies ant-pols as good, suspect, or bad based on the fraction of channels flagged.
+    Flagging takes place in two steps: (1) "channel_diff_flagger" is used to 
+    (2) "dpss_flagger" is used with the array averaged flags to refine initial per-antenna flags
+ 
+    Arguments:
+        data: DataContainer containing antenna autocorrelations (other baselines ignored)
+        good: 2-tuple or list of 2-tuple, default=(0, 0.1)
+            2-tuple or list of 2-tuple bounds for ranges considered good.
+        suspect: 2-tuple or list of 2-tuple, default=(0.1, 0.2)
+            Bounds for ranges considered suspect.
+        nsig: float, default=6
+            The number of sigma in the metric above which to flag pixels. Used in both steps.
+        antenna_class: AntennaClassification, default=None
+            Optional AntennaClassification object. If provided, the flagging method chosen will skip antennas marked "bad".
+            Used in both steps
+        flag_threshold: float, default=0.1
+            The fraction of flags required to trigger a broadcast across all auto-correlations for
+            a given (time, frequency) pixel in the combined flag array. Used in both steps
+        kernel_widths: list
+            Half-width of the convolution kernels used to produce model. True kernel width is (2 * kernel_width + 1)
+            Only used in the "channel_diff_flagger" step
+        mode: str, default='dpss_solve'
+            Method used to solve for DPSS model components. Options are 'dpss_matrix', 'dpss_solve', and 'dpss_leastsq'.
+            Only used in "dpss_flagger" step
+        filter_centers: array-like
+            list of floats of centers of delay filter windows in nanosec. Only used in "dpss_flagger"
+        filter_half_widths: array-like
+            list of floats of half-widths of delay filter windows in nanosec. Only used in "dpss_flagger"
+        
+    Returns:
+        AntennaClassification with "good", "suspect", and "bad" ant-pols based on the absolute fraction of 
+        the band that is flagged
+    """
+    # Flag using convolution kernels
+    antenna_flags, array_flags = xrfi.flag_autos(data, flag_method="channel_diff_flagger", nsig=nsig, antenna_class=antenna_class,
+                                     flag_threshold=flag_threshold, kernel_widths=kernel_widths)
+
+    # Override antenna flags with array-wide flags for next step
+    for key in antenna_flags.keys():
+        antenna_flags[key] = array_flags
+
+    # Flag using DPSS filters
+    antenna_flags, _ = xrfi.flag_autos(data, freqs=data.freqs, flag_method="dpss_flagger", nsig=nsig, antenna_class=antenna_class,
+                                        filter_centers=filter_centers, filter_half_widths=filter_half_widths,
+                                        eigenval_cutoff=eigenval_cutoff, flags=antenna_flags, mode=mode)
+
+    # Calculate the fraction of the band that is flagged
+    flagged_fraction = {bls: np.mean(flags) for bls, flags in antenna_flags.items()}
+
+    return antenna_bounds_checker(flagged_fraction, good=good, suspect=suspect, bad=(-np.inf, np.inf))
--- a/hera_qm/ant_class.py
+++ b/hera_qm/ant_class.py
@@ -321,9 +321,9 @@ def auto_slope_checker(data, good=(-.2, .2), suspect=(-.4, .4), edge_cut=100, fi
 
     return antenna_bounds_checker(relative_slopes, good=good, suspect=suspect, bad=(-np.inf, np.inf))
 
-def auto_rfi_checker(data, good=(0, 0.1), suspect=(0.1, 0.2), nsig=6, antenna_class=None, flag_threshold=0.1, 
-                     kernel_widths=[3, 4, 5], mode='dpss_matrix', filter_centers=[0, -2700e-9, 2700e-9],
-                     filter_half_widths=[200e-9, 200e-9, 200e-9], eigenval_cutoff=[1e-9]):
+def auto_rfi_checker(data, good=(0, 0.1), suspect=(0.1, 0.2), nsig=6, antenna_class=None, flag_broadcast_thresh=0.1, 
+                     kernel_widths=[3, 4, 5], mode='dpss_matrix', filter_centers=[0],
+                     filter_half_widths=[200e-9], eigenval_cutoff=[1e-9]):
     """
     Classifies ant-pols as good, suspect, or bad based on the fraction of channels flagged.
     Flagging takes place in two steps: (1) "channel_diff_flagger" is used to 
@@ -340,18 +340,18 @@ def auto_rfi_checker(data, good=(0, 0.1), suspect=(0.1, 0.2), nsig=6, antenna_cl
         antenna_class: AntennaClassification, default=None
             Optional AntennaClassification object. If provided, the flagging method chosen will skip antennas marked "bad".
             Used in both steps
-        flag_threshold: float, default=0.1
+        flag_broadcast_thresh: float, default=0.1
             The fraction of flags required to trigger a broadcast across all auto-correlations for
             a given (time, frequency) pixel in the combined flag array. Used in both steps
         kernel_widths: list
             Half-width of the convolution kernels used to produce model. True kernel width is (2 * kernel_width + 1)
             Only used in the "channel_diff_flagger" step
-        mode: str, default='dpss_solve'
+        mode: str, default='dpss_matrix'
             Method used to solve for DPSS model components. Options are 'dpss_matrix', 'dpss_solve', and 'dpss_leastsq'.
             Only used in "dpss_flagger" step
-        filter_centers: array-like
+        filter_centers: array-like, default=[0]
             list of floats of centers of delay filter windows in nanosec. Only used in "dpss_flagger"
-        filter_half_widths: array-like
+        filter_half_widths: array-like, default=[200e-9]
             list of floats of half-widths of delay filter windows in nanosec. Only used in "dpss_flagger"
         
     Returns:
@@ -360,7 +360,7 @@ def auto_rfi_checker(data, good=(0, 0.1), suspect=(0.1, 0.2), nsig=6, antenna_cl
     """
     # Flag using convolution kernels
     antenna_flags, array_flags = xrfi.flag_autos(data, flag_method="channel_diff_flagger", nsig=nsig, antenna_class=antenna_class,
-                                     flag_threshold=flag_threshold, kernel_widths=kernel_widths)
+                                     flag_broadcast_thresh=flag_broadcast_thresh, kernel_widths=kernel_widths)
 
     # Override antenna flags with array-wide flags for next step
     for key in antenna_flags.keys():
--- a/hera_qm/ant_class.py
+++ b/hera_qm/ant_class.py
@@ -5,6 +5,7 @@
 """Class and algorithms to classify antennas by various data quality metrics."""
 import numpy as np
 from scipy.ndimage import median_filter
+import warnings
 
 
 def _check_antpol(ap):
@@ -313,7 +314,9 @@ def auto_slope_checker(data, good=(-.2, .2), suspect=(-.4, .4), edge_cut=100, fi
         mean_data = np.mean(data[bl], axis=0)
         med_filt = median_filter(mean_data, size=filt_size)[edge_cut:-edge_cut]
         fit = np.polyfit(np.linspace(-.5, .5, len(mean_data))[edge_cut:-edge_cut], med_filt, 1)
-        relative_slopes[bl] = (fit[0] / fit[1] if np.isfinite(fit[1]) else np.sign(fit[0]) * np.inf)
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore", message="invalid value encountered")
+            relative_slopes[bl] = (fit[0] / fit[1] if np.isfinite(fit[1]) else np.sign(fit[0]) * np.inf)
 
     return antenna_bounds_checker(relative_slopes, good=good, suspect=suspect, bad=(-np.inf, np.inf))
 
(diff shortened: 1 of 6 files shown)


hera_qm/xrfi.py
```diff
@@ -222,6 +224,223 @@ def robust_divide(num, den):
...
+                 suppression_factors=[1e-9], cache={}):
+    """
+    Flag visibilities by filtering data with discrete prolate spheroidal sequences
+
+    Parameters:
+    ----------
+    data: np.ndarray
+        2D data array of the shape (time, frequency)
+    freqs: np.ndarray
+        1D array of frequencies present in the data
```
jsdillon, line 242: Note the proper units of the freqs

hera_qm/xrfi.py
```diff
@@ -222,6 +224,223 @@ def robust_divide(num, den):
...
+    filter_centers: array-like
+        list of floats of centers of delay filter windows in nanosec
+    filter_half_widths: array-like
+        list of floats of half-widths of delay filter windows in nanosec
+    wgts: np.ndarray
+        2D np.ndarray of real weights. Must be the same shape as data.
+    nsig: float, default=10
```
jsdillon, line 249: A default of 10 is different from a lot of the other defaults in this module (usual 5 or 6). Is there a reason for this number?

hera_qm/xrfi.py
```diff
@@ -222,6 +224,223 @@ def robust_divide(num, den):
...
+        suppression_factors=suppression_factors, eigenval_cutoff=eigenval_cutoff,
+        cache=cache
+    )
+    res = data - model
+
+    # Compute modified z-score
+    sigma = np.median(
+        np.abs(res - np.median(res, keepdims=True, axis=1)), axis=1, keepdims=True,
+    )
+    sigma /= 0.675
```
jsdillon, line 294: I'd do 0.6745

hera_qm/xrfi.py
```diff
@@ -222,6 +224,223 @@ def robust_divide(num, den):
...
+    wgts = np.where(np.abs(res) < nsig * sigma, 1, 0)
```
jsdillon, line 295: Why not use the input noise where you have it, instead of 1s?

hera_qm/xrfi.py
```diff
@@ -222,6 +224,223 @@ def robust_divide(num, den):
...
+        List of sinc_matrix eigenvalue cutoffs to use for included DPSS modes.
+    suppression_factors: array-like, default=[1e-9]
+        Specifies the fractional residuals of model to leave in the data. For example, 1e-6 means that the filter
+        will leave in 1e-6 of data fitted by the model.
+    cache: dictionary, default={}
+        dictionary for caching fitting matrices.
+
+    Returns:
+    -------
+    wgts: np.ndarray
```
jsdillon, line 266: It's a bit strange that a function called a "flagger" returns floats and not booleans. It's certainly not what I'd expect from the function name...

hera_qm/xrfi.py
```diff
@@ -222,6 +224,223 @@ def robust_divide(num, den):
...
+
+    # Use noise standard deviation given or estimate it from residuals
+    if noise is not None:
+        sigma = noise
+
+    else:
+        sigma = np.median(
+            np.abs(res - np.median(res, keepdims=True, axis=1)), axis=1, keepdims=True,
+        )
+        sigma /= 0.675
```
jsdillon, line 313: If I'm reading this right, the two cases here produce a signal with different shapes. 

Also, what happens if the noise model also has RFI in it?

hera_qm/xrfi.py
```diff
@@ -222,6 +224,223 @@ def robust_divide(num, den):
...
+    """
+    Flags auto-correlations using channel differencing kernels
+
+    Parameters:
+    ----------
+    data: np.ndarray
+        2D data array of the shape (time, frequency)
+    noise: np.ndarray
+        2D array for containing an estimate of the noise standard deviation. Must be the same shape as the data
+    nsig: float, default=10
```
jsdillon, line 330: Same comment as above

hera_qm/xrfi.py
```diff
@@ -222,6 +224,223 @@ def robust_divide(num, den):
...
+        The number of sigma in the metric above which to flag pixels.
+    kernel_width: list
+        Widths of the convolution kernels used to produce model
```
jsdillon, line 333: These are actually (width - 1)/2, which I'd note explicitly in the docstring

hera_qm/xrfi.py
```diff
@@ -222,6 +224,223 @@ def robust_divide(num, den):
...
+    wgts: np.ndarray, default=None
+        2D array of real weights. Must be the same shape as data.
+
+    Returns:
+    -------
+    wgts: np.ndarray
```
jsdillon, line 339: Same comment as before about bools

hera_qm/xrfi.py
```diff
@@ -222,6 +224,223 @@ def robust_divide(num, den):
...
+
+    # Iterate through differencing kernels of different widths
+    for width in kernel_widths:
+        kernel = np.ones(2 * width + 1)
+        kernel_std = np.sqrt(1 + 1 / (2 * width))
+        dw = data * wgts
+        residual = np.zeros_like(data)
+
+        # For each time integration, identify outliers
+        for t in range(data.shape[0]):
```
jsdillon, line 355: Can't this be done in a vectorized way?

hera_qm/xrfi.py
```diff
@@ -222,6 +224,223 @@ def robust_divide(num, den):
...
+            _d = np.convolve(dw[t], kernel, mode="same") - dw[t]
+            _w = np.convolve(wgts[t], kernel, mode="same") - wgts[t]
+            residual[t] = dw[t] - _d / _w.clip(1, np.inf)
```
jsdillon, line 358: What is the clip doing and what do its bounds mean?

hera_qm/xrfi.py
```diff
@@ -222,6 +224,223 @@ def robust_divide(num, den):
...
+               array_wgts_threshold=0.2, **flag_kwargs):
+    """
+    Driver function for identifying RFI in auto-correlations. Flags auto-correlations on a per antenna basis using
+    a given function and consolidates the weights into a single array-wide flag array.
+
+    Parameters:
+    ----------
+    data: DataContainer
+        DataContainer containing numpy arrays of auto-correlation data with shape (ntimes, nfreqs)
+    flag_method: str
```
jsdillon, line 377: I'd suggest picking a default

hera_qm/xrfi.py
```diff
@@ -222,6 +224,223 @@ def robust_divide(num, den):
...
+            # Estimate the noise standard deviation from data
+            noise = np.abs(v) / np.sqrt(int_count / 2)
+
+            # Run flagging algorithm
+            rfi_wgts[ant1] = alg_func(v, noise=noise, nsig=nsig, wgts=wgts.get(ant1, np.ones_like(v)), **flag_kwargs)
+            array_weights.append((1 - rfi_wgts[ant1]))
+
+    # Combine into array-wide weights
+    array_weights = np.mean(array_weights, axis=0)
+    array_weights = np.where(array_weights < array_wgts_threshold, 1, 0)
```
jsdillon, line 441: If you already know that an antenna is bad, should it count against you here? Like, should the threshold be over all antennas or over the not-already-known-to-be-bad antennas?

jsdillon (changes requested):
Thanks for this @tyler-a-cox! I have a few questions about the overall design, as well as a handful of minor suggestions/clarifications. Wanna chat later today about this?

hera_qm/xrfi.py
tyler-a-cox: Add units to docstring

hera_qm/xrfi.py
tyler-a-cox: This is just what Aaron set in the fastcal notebook for the channel difference flagger. I just changed the default value to `nsig=6` and it didn't significantly change the result on the test data. I'll go ahead and set it to 6

hera_qm/xrfi.py
tyler-a-cox: Changed this.

hera_qm/xrfi.py
tyler-a-cox: Picked the `channel_diff_flagger` as a default.

hera_qm/xrfi.py
tyler-a-cox: Changed to use input noise when possible

hera_qm/xrfi.py
tyler-a-cox: I tried modifying this, but didn't find a solution with a much better speed-up. I can try to modify this again if we notice this getting slow.

hera_qm/xrfi.py
tyler-a-cox: Changed the name of this to `detrend_dpss_filt` to make it more in line with other flaggers in `xrfi` that do similar things

hera_qm/xrfi.py
tyler-a-cox: Removed this

hera_qm/xrfi.py
tyler-a-cox: Modified the name of this similarly to above

hera_qm/xrfi.py
tyler-a-cox: The clip is preventing divide by zero errors. The `_w` array shouldn't have values less than one except at the location of flagged channels, which should have zeros.

tyler-a-cox:
Thanks for all of your comments @jsdillon! This should be ready for another look. I also added a function which classifies antennas by their flag fraction relative to the array-wide flag fraction. I added it to `xrfi`, but a better place for it may be `ant_class`. Let me know what you think!

hera_qm/xrfi.py
jsdillon: I still don't get this. Why? Can't one cook up a set of wgts where that wouldn't be true? 

It seems like maybe this function is written assuming that all weights are 1 or zero, but there's no guarantee of that from the docstring or any kind of error handling.

hera_qm/xrfi.py
```diff
@@ -222,6 +225,266 @@ def robust_divide(num, den):
...
+    return rfi_wgts, array_weights
+
+def auto_rfi_checker(auto_wgts, array_wgts, good=(0, 3), suspect=(3, 6)):
```
jsdillon, line 443: To be consistent with the other function, I think we want this to operate on data directly.

jsdillon: Also, I agree that this belongs in ant_class

hera_qm/xrfi.py
```diff
@@ -222,6 +225,266 @@ def robust_divide(num, den):
...
+    function can be obtained from the "flag_autos" function.
+
+    Parameters:
+    ----------
+    auto_wgts: dictionary or DataContainer
+        Dictionary or DataContainer containing weights for each antenna with values of either zero or one with
+        zero identifying channels found to have RFI
+    array_weights: np.ndarray
+        Contains weights for the entire array averaged over all antennas
+    good: 2-tuple or list of 2-tuple bounds for ranges considered good. In units of relative
```
jsdillon, line 456: Outdated docstrings.

hera_qm/xrfi.py
```diff
@@ -222,6 +225,266 @@ def robust_divide(num, den):
...
+
+    # Compute median and deviation
+    median = np.median(flagged_fraction)
+    std = np.median(np.abs(flagged_fraction - median)) / 0.6745
+
+    # Use standard deviation if std=0
+    if np.isclose(std, 0):
+        std = np.std(flagged_fraction)
+    
+    # For all antennas, determine a z-score
```
jsdillon, line 480: One of the ideas behind ant_class is that we're trying to avoid relative cuts and only doing absolute cuts. 

Instead of z-scores, I think we should either cut on A) absolute fraction of the band flagged or B) absolute fraction of the band in excess of the array-averaged flagging. Perhaps you want to play around with a file and figure out which is better for catching outliers and not flagging good stuff?

jsdillon (changes requested):

hera_qm/xrfi.py
tyler-a-cox: I replaced this with flags instead of weights. I also removed the clipping and went with a robust_divide instead.

hera_qm/xrfi.py
tyler-a-cox: Moved this to ant_class and made the input `data` like the other functions

hera_qm/xrfi.py
tyler-a-cox: Removed this

hera_qm/xrfi.py
tyler-a-cox: I ended up changing this to absolute fraction of the band flagged. It seemed to work fairly well at identifying outlier antennas.

hera_qm/ant_class.py
```diff
@@ -320,6 +321,60 @@ def auto_slope_checker(data, good=(-.2, .2), suspect=(-.4, .4), edge_cut=100, fi
...
+        antenna_class: AntennaClassification, default=None
+            Optional AntennaClassification object. If provided, the flagging method chosen will skip antennas marked "bad".
+            Used in both steps
+        flag_threshold: float, default=0.1
+            The fraction of flags required to trigger a broadcast across all auto-correlations for
+            a given (time, frequency) pixel in the combined flag array. Used in both steps
+        kernel_widths: list
+            Half-width of the convolution kernels used to produce model. True kernel width is (2 * kernel_width + 1)
+            Only used in the "channel_diff_flagger" step
+        mode: str, default='dpss_solve'
```
jsdillon, line 349: Default is out of date

hera_qm/ant_class.py
```diff
@@ -320,6 +321,60 @@ def auto_slope_checker(data, good=(-.2, .2), suspect=(-.4, .4), edge_cut=100, fi
 
     return antenna_bounds_checker(relative_slopes, good=good, suspect=suspect, bad=(-np.inf, np.inf))
 
+def auto_rfi_checker(data, good=(0, 0.1), suspect=(0.1, 0.2), nsig=6, antenna_class=None, flag_threshold=0.1, 
+                     kernel_widths=[3, 4, 5], mode='dpss_matrix', filter_centers=[0, -2700e-9, 2700e-9],
```
jsdillon, line 325: These filter centers are a bit odd and really specialized to the H4C system. Do we actually need to include the 2700 ns feature?

hera_qm/xrfi.py
```diff
@@ -425,65 +420,23 @@ def flag_autos(data, flag_method="detrend_diff_filt", nsig=6, int_count=None, an
...
-            array_weights.append((1 - rfi_wgts[ant1]))
+            rfi_wgts[ant1] = alg_func(
+                v, noise=noise, nsig=nsig, flags=flags.get(ant1, np.zeros_like(v, dtype=bool)), **flag_kwargs
+            )
+            array_weights.append(rfi_wgts[ant1])
 
     # Combine into array-wide weights
     array_weights = np.mean(array_weights, axis=0)
-    array_weights = np.where(array_weights < array_wgts_threshold, 1, 0)
+    array_weights = np.where(array_weights < flag_threshold, False, True)
```
jsdillon, line 437: These aren't really weights anymore, no? Probably the docstring should change too.

hera_qm/xrfi.py
```diff
@@ -328,64 +311,76 @@ def detrend_channel_diff_filt(data, noise, nsig=6, kernel_widths=[1, 2, 4], wgts
...
     int_count: float, default=None
         Number of samples per integration in correlator (dt * dnu)
     antenna_class: AntennaClassification, default=None
         Optional AntennaClassification object. If provided, the flagging method chosen will skip antennas marked "bad"
-    wgts: dictionary or DataContainer: default=None
-        Dictionary or DataContainer containing data weights. If wgts=None, then weights will be set to ones
-    array_wgts_threshold: float, default=0.2
+    flags: dictionary or DataContainer: default={}
+        Dictionary or DataContainer containing boolean flag arrays. If flags=None, then no data will be masked
+    flag_threshold: float, default=0.1
```
jsdillon, line 381: This variable name is a bit misleading, at least to me. I'd call it flag_broadcast_thresh or similar.

hera_qm/xrfi.py
```diff
@@ -276,46 +277,28 @@ def detrend_dpss_filt(data, noise, freqs, filter_centers, filter_half_widths, wg
...
-
+    
     # Determine weights
-    wgts = np.where(np.abs(res) <= nsig * sigma, 1, 0)
-    return wgts
+    return np.where(res > sigma * nsig, True, False)
 
 
-def detrend_channel_diff_filt(data, noise, nsig=6, kernel_widths=[1, 2, 4], wgts=None):
+def channel_diff_flagger(data, noise, nsig=6, kernel_widths=[3, 4, 5], flags=None):
```
jsdillon, line 301: You probably want to update the docstring to clarify that it returns flags, not weights

hera_qm/xrfi.py
```diff
@@ -229,7 +229,7 @@ def robust_divide(num, den):
 # Functions for flagging auto-correlations
 # ############################################################################
 
-def detrend_dpss_filt(data, noise, freqs, filter_centers, filter_half_widths, wgts=None,
+def dpss_flagger(data, noise, freqs, filter_centers, filter_half_widths, flags=None,
```
jsdillon, line 232: Update the docstring returns to reflect that it returns flags not weights

jsdillon (changes requested):
Nice work! Just a handful of documentation fixes and default questions.

hera_qm/ant_class.py
tyler-a-cox: Updated this

hera_qm/ant_class.py
tyler-a-cox: Changed this. We can add those options as input as needed

hera_qm/xrfi.py
tyler-a-cox: Changed the names of these

hera_qm/xrfi.py
tyler-a-cox: Changed the name of this to `flag_broadcast_thresh`

hera_qm/xrfi.py
tyler-a-cox: Updated this

hera_qm/xrfi.py
tyler-a-cox: Updated this

tyler-a-cox:
Thanks for the notes @jsdillon. I think this should be the last time you need to look at this. If I missed anything, let me know!

PR #426, opened 2022-09-19
Use new fast reader on auto_metrics

jsdillon:
This PR uses the new fast reader for auto_metrics, which should speed up i/o a bit.


tyler-a-cox (approved):
This looks good to me! Approved

PR #427, opened 2022-09-26
Save _data on AntennaClassification object created by antenna_bounds_checker for future use

jsdillon:


tyler-a-cox (approved):
This looks good to me.

PR #430, opened 2022-10-07
Slight speedup of `even_odd_zeros_checker()`

jsdillon:
This PR speeds up even_odd_zeros_checker by about 25% by not doing a separate sum calculation for `zero_count_by_ant` which is only used to order antennas.

FYI, just tested the speedup of even_odd_zeros_checker that @tyler-a-cox and @AaronParsons  discussed today.  Unfortunately, it makes it less reliable, because there are many more instances where the data and diff_data have the same real component than cases where both the real and imaginary components match. I’m finding baselines with ~20 matches, which is getting uncomfortably close to the size of an X-engine (making absolute bounds trickier). It may be that those antennas are also broken in some other way, but I’d rather eat the extra second and make this a clean test that clearly reports out which antennas are dropping packets or have broken X-engines.


tyler-a-cox (approved):
This looks good to me

PR #431, opened 2022-10-07
Minor calc_corr_stats() speedup in ant_metrics

jsdillon:
This PR avoids an unnecessary nanmedian speeding up calc_corr_stats() from 11.3s to 9.16s on a pre-loaded file from last night (which is the dominant component of ant_metrics, which takes 13.2 s).

I can't see a way of making calc_corr_stats much faster without vastly increasing the memory footprint by building a giant array of visibilities from many baselines and then doing some clever vectorization of the code. @tyler-a-cox if you have any ideas, I'm all ears.


tyler-a-cox (approved):
Thanks for adding this. I looked into speeding this up too and couldn't find anything. I'll look a little longer, but I'm happy to approve this now.

PR #434, opened 2022-10-14
Sped up ant_metrics.calc_corr_stats and stripped out a lot of complex…

AaronParsons:
…ity in ant_metrics around loading files/baselines iteratively, using flagged data, allowing nans/infs in data, etc. All tests pass.

### DIFF
--- a/hera_qm/ant_metrics.py
+++ b/hera_qm/ant_metrics.py
@@ -307,7 +307,7 @@ class AntennaMetrics():
         for pol in self.same_pols:
             # median over one antenna dimension
             med_corr_matrix = np.nanmedian(self.corr_matrices[pol], axis=0)
-            is_dead = (med_corr_matrix == 0)
+            is_dead = (med_corr_matrix == 0) | ~np.isfinite(med_corr_matrix)
             antpol = self.split_pol(pol)[0]
             dead_ants = [self.ants_per_antpol[antpol][i] for i in np.argwhere(is_dead)[:, 0]]
             for dead_ant in dead_ants:
--- a/hera_qm/ant_metrics.py
+++ b/hera_qm/ant_metrics.py
@@ -66,9 +66,8 @@ def get_ant_metrics_dict():
     return metrics_dict
 
 
-def calc_corr_stats(data_sum, data_diff=None, flags=None, time_alg=np.nanmean, freq_alg=np.nanmean):
-    """Calculate correlation values for all baselines, the average cross-correlation between
-    even and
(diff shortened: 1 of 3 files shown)


hera_qm/ant_metrics.py
```diff
@@ -66,9 +66,8 @@ def get_ant_metrics_dict():
     return metrics_dict
 
 
-def calc_corr_stats(data_sum, data_diff=None, flags=None, time_alg=np.nanmean, freq_alg=np.nanmean):
-    """Calculate correlation values for all baselines, the average cross-correlation between
-    even and
+def calc_corr_stats(data_sum, data_diff=None):
+    """For all baselines, calculate average cross-correlation between even and odd in order to identify ?
```
dstorer, line 70:
```suggestion
    """For all baselines, calculate average cross-correlation between even and odd in order to identify dead, cross-polarized, and non-time-locked antennas.
```

hera_qm/tests/test_ant_metrics.py
```diff
@@ -26,15 +26,10 @@ def test_calc_corr_stats():
...
+    corr_stats = ant_metrics.calc_corr_stats(data_sum,data_diff)
     assert corr_stats[(0,1,'ee')] == 1
     assert corr_stats[(0,1,'nn')] == pytest.approx(0.9578,abs=1e-2)
 
-    # test with flags
-    flags[(0,1,'nn')][1,0] = True
-    corr_stats = ant_metrics.calc_corr_stats(data_sum,data_diff,flags)
-    assert corr_stats[(0,1,'nn')] == pytest.approx(0.9457,abs=1e-2)
-
     # test no diff data
```
dstorer, line 33: We may want to add a test for handling data that are zeros - we probably should have already had a test for this, but I'm particularly concerned now that flag handling has been removed.

dstorer (commented):
Overall this looks good - my only concern is handling of data that's all zeros. Ideally we'd write a test for that scenario.

hera_qm/tests/test_ant_metrics.py
jsdillon: It seems that we broken the special identification of "totally dead" antennas (i.e. median 0 abs(vis)). I've added a new unit test to expose the issue and then I fixed it by undoing one of @AaronParsons's changes

jsdillon (approved):
I've made a few fixes, but now I'm happy with this and I've verified that it produces the same result on a sample file for the file_calibration notebook (zen.2459859.48415.sum.uvh5)

dstorer (approved):
Looks good to me at this point!

PR #435, opened 2022-10-18
Add auto_shape_checker to classify antennas based on their (dis)similarity to the average good autocorrelation

jsdillon:


dannyjacobs (approved):
Looks great to me.

PR #436, opened 2022-11-04
Add handling for channel width attribute in test fixture

bhazelton:
This is future-proofing for a change coming in pyuvdata, but shouldn't break with older versions of pyuvdata.

See this PR for context:
https://github.com/RadioAstronomySoftwareGroup/pyuvdata/pull/1219


jsdillon (approved):
Looks good to me, assuming tests pass.

PR #437, opened 2022-11-10
Enable flagging autos for missing RFI

jsdillon:
This PR enables the flagging of autocorrelations that don't have enough RFI. In practice, this doesn't flag any more antennas (missing RFI usually means other problems) but I think it'll help distinguish FEM off antennas from other issues.


tyler-a-cox (changes requested):
This looks good to me!

PR #438, opened 2022-12-22
Improve ant_metrics handling of zeros

jsdillon:
A recent PR #434 simplified `ant_metrics. calc_corr_stats` but made it less tolerant of zeros appearing in either the odd or even spectrum (or both). As it stands, a single zero in either the even or odd spectrum makes the whole statistic NaN.

If a single X-engine is down, for example, it should still be possible to calculate a corr metric for a given baselines, even if there are some zeros. And in general, we want to do this to help identify white rabbit failures and cross-polarized antennas, which is the purpose of ant_metrics.  This PR robustifies the calculation using a using a nanmean.

If there are zeros in the spectrum, `ant_class. even_odd_zeros_checker` is the right tool to diagnose that. If an antenna is totally dead, we already have a way to handle that in ant_metrics.

This PR adds a test to expose this issue. It also speeds up calc_corr_stats by ~15% by avoiding repeating the multiplication and normalization.

### DIFF
--- a/hera_qm/ant_metrics.py
+++ b/hera_qm/ant_metrics.py
@@ -67,7 +67,10 @@ def get_ant_metrics_dict():
 
 
 def calc_corr_stats(data_sum, data_diff=None):
-    """For all baselines, calculate average cross-correlation between even and odd in order to identify dead, cross-polarized, and non-time-locked antennas.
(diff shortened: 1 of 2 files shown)


Zhilei (approved):
@jsdillon Thanks for the improvement!

PR #439, opened 2023-01-27
Fix some errors with coming pyuvdata changes.

bhazelton:
External hera_qm tests are failing on an upcoming pyuvdata PR because we are adding a warning related to future_array_shapes a long with an option to read object in using the future shape.

~I initially converted the hera_qm code to use the new option whenever a UVData/UVCal/UVFlag object is read in, but then I realized that would only work after the related pyuvdata PR is merged and would require a very recent version of pyuvdata after that. So I changed it to first ensure that hera_qm was always using future shapes (it was in most places but wasn't in a couple places, I figured those were just accidentally overlooked) and then added warnings filters for all the read calls.~

Per the discussion below, this updates the pyuvdata required version to 2.3 and always uses the `future_array_shapes` option when reading files into UVData/UVCal/UVFlag objects. As a part of this update, I also removed some code that is no longer needed now that pyuvdata>=2.3 is required.

I also implement some fixtures to reduce repeated code in `test_xrfi`.


bhazelton:
Looks like there's another pyuvdata PR with hera_qm errors, so I'll update this to pass on that branch as well. I'll also fix the coverage decrease.

bhazelton:
Ok, this now fixes the problems in both pyuvdata PRs.

jsdillon:
I'm OK with requiring the newest version of pyuvdata for the purposes of simplifying this PR (and just not merging it until that version is merged and on pypi).

bhazelton:
@jsdillon Ok, I think this PR is ready. I'm not sure why the codecov check isn't reporting, but it's showing that the patch is 100% covered when you follow the codecov report link.

ci/hera_qm_tests.yml
```diff
@@ -3,15 +3,15 @@ channels:
...
   - conda-forge
 dependencies:
-  - astropy>=3.2.3
+  - astropy>=5.0.4
   - astropy-healpix
-  - h5py
+  - h5py>=3.1
   - matplotlib
-  - numpy>=1.23
+  - numpy>=1.23  # TODO: this is higher than is specified in the setup.py. Should the setup be updated?
```
jsdillon, line 10: yes, probably

hera_qm/tests/test_xrfi.py
```diff
@@ -42,31 +42,232 @@
...
     xrfi.flag_xants(uv, xant)
     assert np.all(uv.flag_array[uv.ant_1_array == xant, :, :])
     assert np.all(uv.flag_array[uv.ant_2_array == xant, :, :])
 
 
-def test_uvcal():
-    uvc = UVCal()
-    uvc.read_calfits(test_c_file)
-    uvc.use_future_array_shapes()
+def test_uvcal(uvcal_calfits):
```
jsdillon, line 262: Mostly out of curiosity, does `uvcal_calfits` have to passed into the function? It's defined as global, right?

jsdillon (dismissed):
One minor question, but looks good to me

ci/hera_qm_tests.yml
bhazelton: ok, I'll change it.

hera_qm/tests/test_xrfi.py
bhazelton: This is how pytest fixtures work.

jsdillon (approved):

PR #440, opened 2023-04-10
Create new module time_series_metrics

jsdillon:
This module brings three functions that currently live in hera_notebook_templates but that I find myself using in multiple notebooks. As such, they should be graduated to proper unit-tested code. There wasn't an obvious home for them, so I felt like this was as good a place as any.

### DIFF
--- a/hera_qm/time_series_metrics.py
+++ b/hera_qm/time_series_metrics.py
@@ -26,7 +26,7 @@ def true_stretches(bool_arr):
 
 
 def impose_max_flag_gap(flags, max_flag_gap=30):
-    '''Adds flags to limit the largest possible number of flagged files between unflagged files.
+    '''Adds flags to limit the largest possible number of flagged times between unflagged times.
     
     Arguments:
         flags_in: 1D boolean numpy array of starting flags (modified in place)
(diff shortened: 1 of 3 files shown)


hera_qm/time_series_metrics.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def metric_convolution_flagging(metric, starting_flags, ok_range, sigma=30, max_flag_gap=30):
+    '''Grows flags by looking at whether the given metric returns to some OK range in the 
+    gap between flags. Also imposes a max_flag_gap (see impose_max_flag_gap()).
+    
+    Arguments:
+        metric: 1D numpy array of floats containing the per-file metric values used for flagging.
+        starting_flags: 1D numpy array of booleans of initial flags
+        ok_range: length-2 tuple of metric range outside of which flags are flags are considered bad
```
tyler-a-cox, line 54: Typo here "flags are flags are"

tyler-a-cox (approved):
One minor typo, but otherwise looks good!

hera_qm/time_series_metrics.py
jsdillon: fixed!

PR #441, opened 2023-05-24
Use the UVCal generic read if available to avoid DeprecationWarnings

bhazelton:
This fixes some errors we're seeing in our most recent pyuvdata runs.


jsdillon (approved):
Looks fine to me!

PR #447, opened 2023-09-07
Future pyuvdata compatibility

bhazelton:
This future proofs for pyuvdata v3.0, which is coming soonish.


jsdillon (approved):
Looks fine to me!

PR #448, opened 2023-09-18
Minor speedup to ant_metrics.calc_corr_stats()

jsdillon:
I've verified that this produces exactly the same result, but improves ant_metrics runtime by ~25% (~12 seconds on the file I tested to ~9 seconds).


tyler-a-cox (approved):
Looks pretty straightforward, approved! Thanks, @jsdillon

PR #453, opened 2024-02-14
Update for uvdata3

steven-murray:
Updates code to work for pyuvdata v3+. Mostly this means assuming that the results from `_key2inds` and `antpair2inds` are possibly slices, rather than arrays of integers. It also fixes a problem with using `pytest.warns(None)` which is deprecated in pytest 8, and replaces it with `pyuvdata.tests.check_warnings()`, which is much simpler anyway.


steven-murray:
@bhazelton note that the missing coverage line will not be missing for pyuvdata 3. It's impossible to cover for pyuvdata 2.

bhazelton (approved):
Looks great, thanks @steven-murray!

PR #454, opened 2024-02-14
ci: add warnings tests and fix some warnings

steven-murray:
This adds a pytest run with warnings that error, so we can catch them sooner rather than later.

I went through and fixed some of the errors that popped up. Some other warnings are fixed in #453, so they will disappear once that is merged. Other warnings I have silenced if they seem reasonable to let go, but I would advise quickly looking over all the "filterwarnings" that I added to make sure that you agree that I'm not hiding anything important.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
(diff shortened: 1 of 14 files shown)


jsdillon (approved):
Looks good to me! I don't think the one warning that's being caught right now is a cause for concern.

PR #455, opened 2024-02-15
maint: switch to using pyproject.toml instead of setup.py/cfg

steven-murray:
This pulls all the code out of `setup.py` and `setup.cfg` and puts it in the python-preferred `pyproject.toml`. 

There should be no change of usage with this PR at all, it's just a more modern way of doing things. 

A couple of things to note:

1. The `pyproject.toml` doesn't support (AFAIK) installing "scripts" the same way as `setup.py`. In fact, it's much nicer, supporting the more general concept of "entry points". How these _should_ be used is to have each "script" as a function that can be run without arguments, and the `pyproject.toml` specifies the CLI command to run that function. This can be built into something really nice if you use a CLI builder like `click`. Here, I haven't done anything nice like that -- I just moved the scripts into the main package (so they get installed) and named their CLI command as the script name (including the `.py` suffix), so there's no change of usage. I would be happy to update this to be a little more modern, so we'd invoke for example: `hera-qm run-xrfi` etc. But I'll leave that up to people who use `hera_qm` more to decide.
2. I added some `ruff` rules to the `pyproject.toml` -- these are simply to reflect the `pycodestyle` rules that were in the `setup.cfg`. Note that these rules weren't being enforced in anyway -- we don't have pre-commit setup on this repo -- so adding them here is not *doing* anything yet. I can make another PR that auto-fixes the rules that aren't ignored.


jsdillon (changes requested):
This looks basically good to me. Two minor requests:

1) Can I request that you do `git mv` on the scripts, rather than deleting them in their old places and creating them in their new ones? It preserves the git blame better and it makes it easier to see what's actually changing when reviewing this PR?

2) Also, can we exclude the scripts from the test coverage percentage?

jsdillon:
Closing this in favor of #456

PR #456, opened 2024-02-15
Use pyprojecttoml redux

steven-murray:
This is the same as #455 but uses `git mv` instead of just moving the files, and removes the scripts from test coverage.


jsdillon (dismissed):
Looks good, though the license is wrong (not that it really matters).

PR #457, opened 2024-02-16
Minimal ruff

steven-murray:
Here I used `adopt-ruff` (https://github.com/ScDor/adopt-ruff) to find all the linting fixes that:

1. We're already obeying
2. Can be automatically fixed by `ruff`
3. We weren't already obeying and aren't auto-fixable but seemed easy enough to fix manually AND important to fix now (e.g. numpy deprecations for numpy v2)

I have *not* run `ruff` format (so no "black" style formatting) and I also ignored line-too-long errors. No functionality should be changed.

I am ignoring one file (`test_xrfi_performance`) which has A LOT of violations, including many "variable name not defined" which makes me think it's completely out of date and should be deleted.

There's one error that I didn't fix because I wasn't sure how to fix it (but it's definitely an error) -- an undefined variable name `i`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,6 @@ hera_qm/data/test_output
 hera_qm/GIT_INFO
 hera_qm/data/test_output/ant_metrics_output.json
 hera_qm/_version.py
+
+artifacts/*
+result.md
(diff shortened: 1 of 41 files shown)


jsdillon:
Looks like we're failing all our tests, @steven-murray mostly related to np.random issues?

jsdillon (approved):

PR #458, opened 2024-03-01
fix: incompatibility with pyuvdata 3

steven-murray:
This fixes an error that arose when testing against pyuvdata 3.0 (seen here: https://dev.azure.com/radioastronomysoftwaregroup/pyuvdata/_build/results?buildId=6079&view=results for example).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,6 +34,7 @@ readme = {file = ["README.md"]}
 tests = [
     "pytest",
     "pytest-cov",
+    "hera-calibration",
 ]
 all = [
     "hera_qm[tests]"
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -34,6 +34,7 @@ readme = {file = ["README.md"]}
 tests = [
     "pytest",
     "pytest-cov",
+    "hera-calibration",
```
jsdillon, line 37: Are you sure you want to do this? Is a circular dependency a concern?

jsdillon (dismissed):
I'm fine with this, though I do have one question.

pyproject.toml
steven-murray: This seems to be fine -- `hera-calibration` is only installed for the `tests` extra, and I tested this on my system and everything works fine. If we don't list it here we need to manually install it when running tests.

jsdillon (approved):

PR #459, opened 2024-05-14
Fixes for compatibility with future pyuvdata

bhazelton:
Fixes a few places for compatibility with the upcoming changes to telescope metadata.


steven-murray (dismissed):
This is all good -- when we require pyuvdata 3 we can remove some lines.

steven-murray (approved):
Thanks a bunch @bhazelton!

PR #461, opened 2024-05-22
Fix warning handling for pyuvdata v3.0

bhazelton:
New warnings are coming in with pyuvdata v3.0. This handles them.


steven-murray (approved):
LGTM. Thanks @bhazelton !

PR #462, opened 2024-05-25
Enable non_noiselike_diff_by_xengine_checker to also use autos

jsdillon:
This fixes a relatively rare problem in RTP where flagging doesn't completely catch every antenna with missing packets because the strongest evidence for problems showed up in the autos (which was excluded for simplicity before). This corrects that oversight.

Here's an example from `zen.2460451.16874.sum.uvh5`:

image

Before `xengine_diff_class`

image

After: 

image

With fix:

image

I've verified that there's no effect on `zen.2460451.34457.sum.uvh5` which was fine.


tyler-a-cox (approved):
This looks okay to me! Thanks, @jsdillon!

PR #463, opened 2024-06-13
Updates for pyuvdata v3

bhazelton:
In version 3, the import for pyuvdata's warning_checker function will be different, this handles that change.


steven-murray (approved):
Looks good, thanks @bhazelton

PR #465, opened 2025-02-07
Fix slope checker when one ant-pol is completely flagged

jsdillon:
One one pol is flagged, this used to raise an error (which was a problem in file_calibration, as well as the auto_checker). This should fix that.


tyler-a-cox (approved):
This looks good to me!

